PR #19, opened 2020-06-13
lifetime scope for cached IDisposable

plexcake:
Implement a wrapper class for IDisposable so that it can safely be stored in a cache that will dispose the object when it is removed.


Lightweight.Caching.UnitTests/ScopedDisposableTests.cs
```diff
@@ -0,0 +1,64 @@
...
+			scope.Dispose();
+
+			scope.Invoking(s => s.CreateLifetime()).Should().Throw<ObjectDisposedException>();
+		}
+
+        private class Disposable : IDisposable
+        {
+			public bool IsDisposed { get; set; }
+
+            public void Dispose()
```
plexcake, line 56: Formatting

Lightweight.Caching.UnitTests/ScopedTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            var disposable = new Disposable();
+            var scope = new Scoped<Disposable>(disposable);
+            scope.Dispose();
+
+            scope.Invoking(s => s.CreateLifetime()).Should().Throw<ObjectDisposedException>();
+        }
+
+        [Fact]
+        public void WhenScopeIsCreatedFromCacheLifetimeCanBeCreatedAndDisposed()
```
plexcake, line 54: WhenLifetimeIsCreatedFromCacheItemHasExpectedLifetime

PR #20, opened 2020-06-13
fix remove

bitfaster:


Lightweight.Caching/Lru/TemplateConcurrentLru.cs
```diff
@@ -290,11 +292,18 @@ private void Move(I item, ItemDestination where)
...
                     break;
                 case ItemDestination.Remove:
-                    if (this.dictionary.TryRemove(item.Key, out var removedItem))
-                    {
-                        if (removedItem.Value is IDisposable d)
+                    if (!item.WasRemoved)
+                    {       
+                        // We only mark the item that was actually removed as removed. ConcurrentDictionary.TryRemove 
+                        // is thread sage, so only 1 thread can remove each item. If there is a race with this.TryRemove
+                        // only 1 thread will actually remove the item.
```
peck-a, line 299: How about this scenario with the same key:
1. Thread A TryRemove, removes item1, has reference to removed item
2. Thread B GetOrAdd => Adds item2
3. Thread C Cycle => move(Removed), has reference to item 1

Can thread C remove thread A's item?

peck-a: If TValue is a reference type, it is possible to detect this situation in Thread C by reference comparing item and removedItem.

How could we recover?

peck-a: Issue is that LruItem1* is present in the queues somewhere, but is not in the dictionary.

1. Sweep all queues until we find the matching reference, mark it as removed.
2. Add to the dictionary with ConcurrentDictionary.TryAdd(removedItem) - this could be another race
3. Store the value in a list of inconsistencies and use the inconsistency list to recover

bitfaster: Fixed by making TryRemove first fetch the item in order to obtain the lock.

PR #47, opened 2020-06-29
Atomic value creation

bitfaster:
By caching Lazy, we can easily guarantee a single invocation of the value create logic. However, the standard Lazy implementation has some drawbacks:

1. It caches exceptions, and there is no way turn this off. This isn't how ConcurrentDictionary.GetOrAdd works (only values that are able to be created are stored).
2. The Func value factory is cached until the value is created or fails to create. This means that even if problem 1 were solved, in the case where the factory relied on some 'per invocation' state like a local connection used to query a database, subsequent invocations from other threads could fail if the database connection was closed.
3. It does [not handle IDisposable](https://stackoverflow.com/questions/50935483/thread-safe-disposal-of-lazy-initialized-objects) values.
4. There is no built in AsyncLazy.

This PR introduces a family of Atomic classes that guarantee atomic value creation, do not cache exceptions, handle IDisposable, async value creation and async creation of IDisposable values.


BitFaster.Caching/LazyExperiments/ScopedAsyncLazy.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace BitFaster.Caching.LazyExperiments
+{
+    // Enable caching an AsyncLazy disposable object - guarantee single instance, safe disposal
+    public class ScopedAsyncLazy<TValue> : IDisposable 
+        where TValue : IDisposable
+    {
+        private ReferenceCount<Lazy<TValue>> refCount;
```
bitfaster, line 13: This is never initialized.

BitFaster.Caching/Lazy/Atomic.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace BitFaster.Caching.Lazy
+{
+    // Should this be called simply Atomic?
+    // Then we have Atomic and Scoped, and ScopedAtomic
+    // Then AtomicAsync, ScopedAsync, and ScopedAtomicAsync would follow, but there is no ScopedAsync equivalent at this point. That should be IAsyncDisposable (.net Core 3.1 onwards). That would imply that GetOrAdd async understands IAsyncDispose
+
+    // https://github.com/dotnet/runtime/issues/27421
+    // https://github.com/alastairtree/LazyCache/issues/73
+    public class Atomic<T>
```
bitfaster, line 17: Compare to Lazy:
https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Private.CoreLib/src/System/Lazy.cs#L319

- Lazy stores only value, func, and a state object (size = 3 refs)
- This has 3 reference type members plus isInitialized
- value factory is not set to null when done - could leak the factory

BitFaster.Caching/Lazy/Atomic.cs
```diff
@@ -0,0 +1,72 @@
...
+        private bool isInitialized;
+
+        private object @lock;
+
+        public Atomic(Func<T> factory)
+        {
+            valueFactory = factory;
+        }
+
+        public T Value => LazyInitializer.EnsureInitialized(ref value, ref isInitialized, ref @lock, valueFactory);
```
bitfaster, line 32: Need some special handling of this in the debugger, so that the debugger doesn't cause the lazy to become initialized.

BitFaster.Caching/Lazy/Atomic.cs
```diff
@@ -0,0 +1,72 @@
...
+            try
+            {
+                return await LazyInitializer.EnsureInitialized(ref task, ref isInitialized, ref @lock, valueFactory).ConfigureAwait(false);
+            }
+            catch
+            {
+                Volatile.Write(ref isInitialized, false);
+                throw;
+            }
+        }
```
bitfaster, line 63: Let's say we create a private initializer class that holds a reference to the factory, the lock etc.

```cs
public async Task<T> Value()
{
   if (this.initializer == null)
   {
       return this.value;
   }


   this.value= await this.initializer.CreateAsync().ConfigureAwait(false);
   this.initializer = null;
   return value;
```

BitFaster.Caching/Lazy/DisposableAtomic.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Threading.Tasks;
+
+namespace BitFaster.Caching.Lazy
+{
+    // TODO: is this actually even needed? Or is the approach in ScopedAsyncAtomic sufficient? E.g. rely on IsValueCreated at dispose time, scoped owns tracking dispose and is already thread safe.
+    // requirements for IDisposable atomic
+    // if value !created, no dispose, cannot create - throws object disposed exception
+    // if created, dispose value
+    [DebuggerDisplay("IsValueCreated={IsValueCreated}, Value={ValueIfCreated}")]
+    public class DisposableAtomic<K, V> : IDisposable where V : IDisposable
```
bitfaster, line 16: Could also create a ScopedLazy, allowing the user to substitute in Lazy behavior if preferred to Atomic.

This would work the same way - check the value is created property when disposing.

BitFaster.Caching/Lazy/AsyncAtomic.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace BitFaster.Caching
+{
+    [DebuggerDisplay("IsValueCreated={IsValueCreated}, Value={ValueIfCreated}")]
+    public class AsyncAtomic<K, V>
```
bitfaster, line 12: This would support both sync and async.

Sync adds are slightly penalized vs pure Lazy because `Initializer.CreateValue` creates a TCS, which internally creates a Task<V>. These are both pretty lightweight - likely cheap compared to value factory.

```C#
namespace BitFaster.Caching
{
    [DebuggerDisplay("IsValueCreated={IsValueCreated}, Value={ValueIfCreated}")]
    public class AsyncAtomic<K, V>
    {
        private Initializer initializer;

        [DebuggerBrowsable(DebuggerBrowsableState.Never)]
        private V value;

        public AsyncAtomic()
        {
            this.initializer = new Initializer();
        }

        public AsyncAtomic(V value)
        {
            this.value = value;
        }

        public V GetValue(K key, Func<K, V> valueFactory)
        {
            if (this.initializer == null)
            {
                return this.value;
            }

            return CreateValue(key, valueFactory);
        }

        public async Task<V> GetValueAsync(K key, Func<K, Task<V>> valueFactory)
        {
            if (this.initializer == null)
            {
                return this.value;
            }

            return await CreateValueAsync(key, valueFactory).ConfigureAwait(false);
        }

        public bool IsValueCreated => this.initializer == null;

        public V ValueIfCreated
        {
            get
            {
                if (!this.IsValueCreated)
                {
                    return default;
                }

                return this.value;
            }
        }

        private V CreateValue(K key, Func<K, V> valueFactory)
        {
            Initializer init = this.initializer;

            if (init != null)
            {
                this.value = init.CreateValue(key, valueFactory);
                this.initializer = null;
            }

            return this.value;
        }

        private async Task<V> CreateValueAsync(K key, Func<K, Task<V>> valueFactory)
        {
            Initializer init = this.initializer;

            if (init != null)
            {
                this.value = await init.CreateValueAsync(key, valueFactory).ConfigureAwait(false);
                this.initializer = null;
            }

            return this.value;
        }

        private class Initializer
        {
            private object syncLock = new object();
            private bool isInitialized;
            private Task<V> valueTask;

            public V CreateValue(K key, Func<K, V> valueFactory)
            {
                var tcs = new TaskCompletionSource<V>(TaskCreationOptions.RunContinuationsAsynchronously);

                var synchronizedTask = Synchronized.Initialize(ref this.valueTask, ref isInitialized, ref syncLock, tcs.Task);

                if (ReferenceEquals(synchronizedTask, tcs.Task))
                {
                    try
                    {
                        var value = valueFactory(key);
                        tcs.SetResult(value);
                        return value;
                    }
                    catch (Exception ex)
                    {
                        Volatile.Write(ref isInitialized, false);
                        tcs.SetException(ex);
                        throw;
                    }
                }

                // TODO: how dangerous is this?
                // it can block forever if value factory blocks
                return synchronizedTask.GetAwaiter().GetResult();
            }

            public async Task<V> CreateValueAsync(K key, Func<K, Task<V>> valueFactory)
            {
                var tcs = new TaskCompletionSource<V>(TaskCreationOptions.RunContinuationsAsynchronously);

                var synchronizedTask = Synchronized.Initialize(ref this.valueTask, ref isInitialized, ref syncLock, tcs.Task);

                if (ReferenceEquals(synchronizedTask, tcs.Task))
                {
                    try
                    {
                        var value = await valueFactory(key).ConfigureAwait(false);
                        tcs.SetResult(value);

                        return value;
                    }
                    catch (Exception ex)
                    {
                        Volatile.Write(ref isInitialized, false);
                        tcs.SetException(ex);
                        throw;
                    }
                }

                return await synchronizedTask.ConfigureAwait(false);
            }
        }
    }

    internal static class Synchronized
    {
        public static V Initialize<V>(ref V target, ref bool initialized, ref object syncLock, V value)
        {
            // Fast path
            if (Volatile.Read(ref initialized))
            {
                return target;
            }

            lock (syncLock)
            {
                if (!Volatile.Read(ref initialized))
                {
                    target = value;
                    Volatile.Write(ref initialized, true);
                }
            }

            return target;
        }
    }
}
```

bitfaster: Then use a decorator:

```C#
namespace BitFaster.Caching
{
    public class AtomicCacheDecorator<K, V> : ICache<K, V>
    {
        private readonly ICache<K, AsyncAtomic<K, V>> cache;

        public AtomicCacheDecorator(ICache<K, AsyncAtomic<K, V>> cache)
        {
            this.cache = cache;
        }

        public int Capacity => this.cache.Capacity;

        public int Count => this.cache.Count;

        public void AddOrUpdate(K key, V value)
        {
            throw new NotImplementedException();
        }

        public void Clear()
        {
            this.cache.Clear();
        }

        public V GetOrAdd(K key, Func<K, V> valueFactory)
        {
            var synchronized = cache.GetOrAdd(key, _ => new AsyncAtomic<K, V>());
            return synchronized.GetValue(key, valueFactory);
        }

        public Task<V> GetOrAddAsync(K key, Func<K, Task<V>> valueFactory)
        {
            var synchronized = cache.GetOrAdd(key, _ => new AsyncAtomic<K, V>());
            return synchronized.GetValueAsync(key, valueFactory);
        }

        public void Trim(int itemCount)
        {
            this.cache.Trim(itemCount);
        }

        public bool TryGet(K key, out V value)
        {
            AsyncAtomic<K, V> output;
            bool ret = cache.TryGet(key, out output);

            if (ret && output.IsValueCreated)
            {
                value = output.ValueIfCreated;
                return true;
            }

            value = default;
            return false;
        }

        public bool TryRemove(K key)
        {
            return this.cache.TryRemove(key);
        }

        public bool TryUpdate(K key, V value)
        {
            return cache.TryUpdate(key, new AsyncAtomic<K, V>(value)); ;
        }
    }
}
```

bitfaster:
Replaced with the following PRs to support atomic create: https://github.com/bitfaster/BitFaster.Caching/pull/155 + https://github.com/bitfaster/BitFaster.Caching/pull/153

bitfaster:
Replaced with the following PRs to support atomic create: https://github.com/bitfaster/BitFaster.Caching/pull/155 + https://github.com/bitfaster/BitFaster.Caching/pull/153

PR #69, opened 2021-05-31
Provide extension methods to create Lifetime<T> for ICache Scoped<T>

bitfaster:
Extension methods to pull values from the cache when they are scoped, automatically retrying if the scope cannot be created because there was a race causing the scoped object to be disposed just as it was returned (by a race in cache eviction).

This could also be implemented as a decorator on top of ICache, so that all of the methods transparently transform to scope/lifetime (e.g. TryUpdate would scope input arg).

Should Lifetime be a struct? Could rely on the IsDisposed flag from the inner RefCount object. This would eliminate a heap alloc, and would be immutable since it only holds pointers.

### DIFF
--- a/BitFaster.Caching/ScopedCache.cs
+++ b/BitFaster.Caching/ScopedCache.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BitFaster.Caching
+{
+    // what happens if we completely encapsulate scoped?
+    // we can't implement ICache, since return types will now be Lifetime, not T
+    public class ScopedCache<K, T> where T : IDisposable
+    {
+        private readonly ICache<K, Scoped<T>> innerCache;
+
+        public ScopedCache(ICache<K, Scoped<T>> innerCache)
+        {
+            this.innerCache = innerCache;
+        }
+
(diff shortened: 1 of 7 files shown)


BitFaster.Caching/ScopedCache.cs
```diff
@@ -0,0 +1,78 @@
...
+                    return true;
+                }
+            }
+
+            value = default(Lifetime<T>);
+            return false;
+        }
+
+        // If we completely encapsulate Scoped, then we must allocate a new value factory for every GetOrAdd call
+        public Lifetime<T> GetOrAdd(K key, Func<K, Scoped<T>> valueFactory)
```
bitfaster, line 36: This isn't totally fool proof - if the caller's delegate returns a Scoped instance that is disposed, the GetOrAdd methods will infinite loop.

There is no obvious way to check this, since the delegate is called by ICache, so ScopedCache can't see it. By the time ScopedCache sees the value, it has no way to differentiate between a race where Scoped got disposed, and a deterministically broken value factory.

bitfaster: Options for safe scopes are:

1. Let the user screw it up, but have best perf
2. For debug only, insert an additional layer and verify new scoped instances are not disposed
3. Always insert the additional layer, suck up the extra delegate allocs
4. Force user to implement a factory class, factory enforces the constraint

bitfaster:
Superseded by https://github.com/bitfaster/BitFaster.Caching/pull/135

PR #76, opened 2021-08-21
Target .NET 6

bitfaster:
Add build target for .NET 6.


bitfaster:
Currently CodeQL analysis fails in Autobuild.

Issue #81, opened 2021-10-19
Add a GetOrAdd that takes a value?


mburbea:
ConcurrentDictionary has an overload of GetOrAdd that takes a value instead of a valueFactory, can an overload be added for the Lru?

bitfaster:
I have been working on some ideas for caching Lazy values that would behave similarly to use of ConcurrentDictionary.GetOrAdd with respect to exceptions, i.e. not to cache them so the program isn't stuck in a terminal state. This is the [PR](https://github.com/bitfaster/BitFaster.Caching/pull/47).

I combined your approach into the class AsyncAtomic, which lets you write code as below with the same GetOrAddAsync signature (via an extension method):

```cs
var asyncAtomicLru = new ConcurrentLru<int, AsyncAtomic<int, int>>(5);

int ar = await asyncAtomicLru.GetOrAddAsync(1, i => Task.FromResult(i));
```

In a 100% cache hit benchmark, the lookup performance of this wrapper is about 20% slower than ConcurrentLru.GetOrAddAsync with no memory allocs - no delegate alloc (numbers below based on .NET 6). This is a relatively small penalty to synchronize creation - it's about half the overhead compared to your code (I copy pasted it into the benchmark, you can see below).

 |                   Method |      Mean | Ratio | Code Size | Allocated |
|------------------------- |----------:|------:|----------:|----------:|
|     ConcurrentDictionary.GetOrAdd |  7.096 ns |  0.16 |     340 B |         - |
|       ConcurrentLru.GetOrAddAsync | 45.457 ns |  1.00 |     495 B |         - |
| AtomicConcurrentLru.GetOrAddAsync | 55.937 ns |  1.23 |     495 B |         - |
|     mburbea.GetOrAddAsync | 64.890 ns |   1.43 |     495 B |       24 B |

This is the benchmark code:

```cs
    [DisassemblyDiagnoser(printSource: true)]
    [MemoryDiagnoser]
    public class AsyncAtomicBench
    {
        private static readonly ConcurrentDictionary<int, int> dictionary = new ConcurrentDictionary<int, int>(8, 9, EqualityComparer<int>.Default);

        private static readonly ConcurrentLru<int, int> concurrentLru = new ConcurrentLru<int, int>(8, 9, EqualityComparer<int>.Default);

[...]

PR #82, opened 2021-10-24
Synchronized value factory

bitfaster:
Based on [this](https://github.com/bitfaster/BitFaster.Caching/issues/81), enable the following:

```cs
var lru = new ConcurrentLru<int, string>(9);
Func<int, string> valueFactory = x => x.ToString();
var synchronized = new SynchronizedValueFactory<int, string>(lru, valueFactory);

var res = lru.GetOrAdd(1, synchronized.Create);
```

Could this be an additional policy class that is part of the template?


BitFaster.Caching/SynchronizedValueFactory.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public V Create(K key)
+        {
+            using (var sema = this.semaphoreCache.Acquire(key, k => new SemaphoreSlim(1, 1)))
+            {
+                sema.Value.Wait();
+
+                try
+                {
+                    return this.cache.GetOrAdd(key, this.valueFactory);
```
bitfaster, line 46: This is not correct, since in the GetOrAdd method LRU will find it can't add this value, then try to dispose it if it is IDisposable.

This would result in there being a disposed value in the cache. So probably here we should return default(T).

bitfaster:
This approach is not functionally correct, see comments.

PR #90, opened 2021-11-10
Add event when item is evicted

bitfaster:
Modify HitCounter to expose OnItemRemoved() and plumb into an event on ConcurrentLru/ConcurrentTLru. Event fires with a reason (either removed or evicted) whenever an item is removed. Event is invoked before value is disposed (if it is IDisposable).

Enables this code to be written:

```
var lruEvents = new ConcurrentLru<int, int>(1, 6, EqualityComparer<int>.Default);
List<ItemRemovedEventArgs<int, int>> removedItemList = new();

lruEvents.ItemRemoved += (source, eventArgs) => removedItemList.Add(eventArgs);
lruEvents.GetOrAdd(1, k => k);
lruEventsTryRemove(1);
// event will fire!
```

LRU.GetOrAdd method total bytes of assembly code:

Fast (event omitted) 280 bytes
Event not registered: 306 bytes
Event registered: 306 bytes

Compared to main, GetOrAdd is increased from 302 to 306 bytes:

image

LRU.Move method total bytes of assembly code:

Fast (event omitted) 441 bytes
Event not registered: 508 bytes
Event registered: 508 bytes

image

Making the event have an empty body to avoid the null check (which is only possible in .NET 6 because it requires a struct field initializer), makes the code 503 bytes but then the event arg is always allocated even if there is no subscribed listener.


BitFaster.Caching.UnitTests/Lru/ConcurrentLruTests.cs
```diff
@@ -359,6 +359,29 @@ public void WhenValueExpiresItIsDisposed()
...
+            var lruEvents = new ConcurrentLru<int, int>(1, 6, EqualityComparer<int>.Default);
+            lruEvents.ItemRemoved += OnLruItemRemoved;
+
+            for (int i = 0; i < 5; i++)
+            {
+                lruEvents.GetOrAdd(i, i => i);
+            }
+
+            removedItems.Count.Should().Be(1);
+            removedItems[0].Key.Should().Be(0);
```
bitfaster, line 374: this is kind of a crappy test because default value of int is zero - change to reflect some values that would detect a failure. Also need equivalent test on TLru.

PR #94, opened 2021-11-13
Build wrapped value into ItemPolicy

bitfaster:
It is awkward to handle caches of Scoped<V> - either need a wrapper factory Func<> to convert which is always a closure (alloc), or user can pass in Scoped, which opens the door to creating a Disposed scoped object which can put the cache into an invalid state. See this [PR](https://github.com/bitfaster/BitFaster.Caching/pull/69) for impl.

Now make policy have a type param W for the wrapped type, which can be Scoped.

- Problem remains that we return Scoped, and therefore enable the user to dispose scopes still in the cache and corrupt it (since a non removed item now cannot create lifetimes)
- Then introduce type param R, return type of GetOrAdd
- Policy has a method to convert W into R, which would create lifetime for Scoped, and just return the value for non wrapped types which presumably the JIT can completely elide. 

Does this extend to Atomic etc. ? Would need to extend policy so that CreateItem accepts factory and to have an init method that accepts the factory. Then inside policy choose when factory is called. Question here is whether JIT will eliminate unnecessary args for specialized policies (factory is invoked at one call site or the other).

### DIFF
--- a/BitFaster.Caching/ICache.cs
+++ b/BitFaster.Caching/ICache.cs
@@ -6,7 +6,7 @@ using System.Threading.Tasks;
 
 namespace BitFaster.Caching
 {
-    // can have this as internal, and inherit ICache<K, V> from ICacheBase<K, V, V> and IScopedCache : ICacheBase<K, V, Scoped<V>>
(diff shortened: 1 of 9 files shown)


bitfaster:
Abandoned in favor of ScopedCache.

PR #125, opened 2022-07-09
build async+atomic+scoped

bitfaster:
Atomic builders do not support mixing sync and async (at least not in a way that would be safe if the caller mixes sync/async calls). Therefore, segregate the factory methods on `ICache` and `IScopedCache` so that sync and async calls cannot be mixed:

```cs
interface ICache { V GetOrAdd(K key, Func<K, V> valueFactory); }
interface IAsyncCache { Task<V> GetOrAddAsync(K key, Func<K, Task<V>> valueFactory); }
interface IScopedCache { Lifetime<V> ScopedGetOrAdd(K key, Func<K, Scoped<V>> valueFactory); }
interface IScopedAsyncCache { Task<Lifetime<V>> ScopedGetOrAddAsync(K key, Func<K, Task<Scoped<V>>> valueFactory); }
```

Extend LRU builder to support all combinations of:

- Metrics
- Expire after write (TLru)
- Scoped IDisposable values
- Synchronized factory method
- Async factory method

Use of builder now looks like this (methods can be called in any order with the same result):

```cs
var cache = new LruBuilder<int, Disposable>()
    .WithCapacity(666)
    .WithKeyComparer(EqualityComparer<int>.Default)
    .WithConcurrencyLevel(4)
    .WithAtomicCreate()
    .WithMetrics()
    .WithScopedValues()
    .AsAsyncCache()
    .Build();
```


BitFaster.Caching.UnitTests/Lru/LruBuilderTests.cs
```diff
@@ -86,5 +122,25 @@ public void TestConcurrencyLevel()
...
             constructor.Should().Throw<ArgumentOutOfRangeException>();
         }
+
+        [Fact]
+        public void ScopedPOC()
+        {
+            // Choose from 16 combinations of Lru/TLru, Instrumented/NotInstrumented, Atomic create/not atomic create, scoped/not scoped
+
+            // layer 1: can choose ConcurrentLru/TLru, FastConcurrentLru/FastConcurrentTLru 
+            var c = new ConcurrentLru<int, AsyncAtomic<int, Scoped<Disposable>>>(3);
```
bitfaster, line 132: This doesn't work for IDisposable.

Cache sees AsyncAtomic, which is not IDisposable - thus the inner scoped item will never be disposed.

There should probably be DisposableAtom. Somehow this must handle races between IsValueCreated and Dispose.

BitFaster.Caching/Synchronized/IdempotentAsyncCache.cs
```diff
@@ -0,0 +1,78 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BitFaster.Caching.Synchronized
+{
+    public class IdempotentAsyncCache<K, V> : ICache<K, V>
```
bitfaster, line 9: AtomicFactoryCache
AtomicFactoryAsyncCache
AtomicFactoryScopedCache
AtomicFactoryScopedAsyncCache

PR #135, opened 2022-07-17
scopedec

bitfaster:
Provide an explicit ScopedCache to ease working with scoped IDisposable values. All read operations are prefixed with 'Scoped' and return lifetimes instead of values. Scoped methods implement logic to ensure the lifetime returned is always valid.

```cs
public class Foo
{
    private const int capacity = 666;
    private ScopedCache<int, SomeDisposable> lru = new(new ConcurrentLru<int, Scoped<SomeDisposable>>(capacity));

    public void Bar()
    {
        var valueFactory = new SomeDisposableValueFactory();

        using (var lifetime = lru.ScopedGetOrAdd(1, k => valueFactory.Create(k)))
        {
            // lifetime.Value is guaranteed to be alive until the lifetime is disposed
        }
    }
}

public class SomeDisposableValueFactory
{
    public Scoped<SomeDisposable> Create(int key)
    {
        return new Scoped<SomeDisposable>(new SomeDisposable(key));
    }
}
```

### DIFF
--- a/BitFaster.Caching/IScopedCache.cs
+++ b/BitFaster.Caching/IScopedCache.cs
@@ -6,6 +6,11 @@ using System.Threading.Tasks;
 
 namespace BitFaster.Caching
 {
+    /// <summary>
+    /// Represents a generic cache of key/scoped IDisposable value pairs.
+    /// </summary>
+    /// <typeparam name="K">The type of keys in the cache.</typeparam>
+    /// <typeparam name="V">The type of values in the cache.</typeparam>
     public interface IScopedCache<K, V> where V : IDisposable
     {
         /// <summary>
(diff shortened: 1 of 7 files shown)


BitFaster.Caching/IScopedCache.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        int Count { get; }
+
+        /// <summary>
+        /// Attempts to create a lifetime for the value associated with the specified key from the cache
+        /// </summary>
+        /// <param name="key">The key of the value to get.</param>
+        /// <param name="lifetime">When this method returns, contains a lifetime for the object from the cache that has the specified key, or the default value of the type if the operation failed.</param>
+        /// <returns>true if the key was found in the cache; otherwise, false.</returns>
+        bool ScopedTryGet(K key, out Lifetime<V> lifetime);
```
bitfaster, line 32: This is kind of a bad API because its cumbersome to wrap lifetime in a using statement.

BitFaster.Caching/ScopedCache.cs
```diff
@@ -0,0 +1,124 @@
...
+                if (scope.TryCreateLifetime(out var lifetime))
+                {
+                    return lifetime;
+                }
+
+                spinwait.SpinOnce();
+
+                if (c++ > MaxRetry)
+                {
+                    throw new InvalidOperationException();
```
bitfaster, line 86: What exception should this be?

BitFaster.Caching/IScopedCache.cs
bitfaster: This can be provided as an extension method.

PR #139, opened 2022-07-17
events

bitfaster:
Follow metrics pattern for events. Later there will be more events (e.g. ItemUpdated etc.), so good to extract this out.

There is a kind of nasty part here due to the telemetry policy being a struct (in order to get the JIT optimizations). If the struct contains an event, += or -= an event handler when the policy has been returned directly seems to cause the struct to be copied.  This doesn't affect the wrapper event in ConcurrentLru/ConcurrentTLru which did the same thing without going through the Events property (these are now updated to go through the Events property).

Therefore, put an 'EventHolder' class inside the struct, and initialize it when the event source is set. It might be better in the long run to eliminate the event from the policy and have the event holder as an explicit class passed into the ctor somehow. That way the struct goes back to only value types/source object.

Perf evaluation: JustGetOrAdd test is as expected - adding the class within the policy has not affected JIT (else 'Fast' latency would match 'non-Fast' variants).

### DIFF
--- a/BitFaster.Caching/IScopedCache.cs
+++ b/BitFaster.Caching/IScopedCache.cs
@@ -31,29 +31,35 @@ namespace BitFaster.Caching
         /// <summary>
         /// Gets the cache events.
         /// </summary>
+        /// <remarks>
+        /// Events expose the Scoped instance wrapping each value. To keep the value alive (blocking Dispose), try to 
+        /// create a Lifetime from the scope.
+        /// </remarks>
         ICacheEvents<K, Scoped<V>> Events { get; }
 
         /// <summary>
         /// Attempts to create a lifetime for the value associated with the specified key from the cache
         /// </summary>
         /// <param name="key">The key of the value to get.</param>
-        /// <param name="lifetime">When this method returns, contains a lifetime for the object from the cache that has the specified key, or the default value of the type if the operation failed.</param>
+        /// <param name="lifetime">When this method returns, contains a lifetime for the object from the cache that 
+        /// has the specified key, or the default value of the type if the operation failed.</param>
         /// <returns>true if the key was found in the cache; otherwise, false.</returns>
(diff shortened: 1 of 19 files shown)


BitFaster.Caching/IScopedCache.cs
```diff
@@ -28,6 +28,11 @@ public interface IScopedCache<K, V> where V : IDisposable
         /// </summary>
         ICacheMetrics Metrics { get; }
 
+        /// <summary>
+        /// Gets the cache events.
+        /// </summary>
+        ICacheEvents<K, Scoped<V>> Events { get; }
```
bitfaster, line 34: This exposes the scope object to the outside world, enabling bad things to happen (e.g. caller disposes the scope, now the cache contains a disposed object).

Alternative would be to make some sort of wrapper to proxy the event and pull the object out, sending a potentially disposed object to the event handler.

It seems better to send the scope. The caller can then create a lifetime if they need to work on the value, else ignore it.

BitFaster.Caching/Lru/ConcurrentTLru.cs
```diff
@@ -55,6 +55,7 @@ public ConcurrentTLru(int concurrencyLevel, ICapacityPartition capacity, IEquali
         /// <summary>
         /// Occurs when an item is removed from the cache.
         /// </summary>
+        [ObsoleteAttribute("This property is obsolete. Use Events instead.", false)]
         public event EventHandler<ItemRemovedEventArgs<K, V>> ItemRemoved
         {
             add { this.telemetryPolicy.ItemRemoved += value; }
```
bitfaster, line 61: Should switch these to go through ICache.Events, then all tests are equivalent.

BitFaster.Caching/IScopedCache.cs
bitfaster: Added comment.

PR #153, opened 2022-07-21
idempotent value creation

bitfaster:
Provide building blocks to synchronize cache value factories in order to mitigate a [cache stampede](https://en.wikipedia.org/wiki/Cache_stampede).

In each case below, multiple threads calling GetValue equivalents will be synchronized such that only a single thread runs at a time and the others see the result of the winning thread:

```cs
var idempotent = new Idempotent<int, int>();
int i = idempotent.GetValue(1, k => k);
```

```cs
var idempotent = new AsyncIdempotent<int, int>();
int i = await idempotent.GetValueAsync(1, k => Task.FromResult(k));
```

```cs
var idempotent = new ScopedIdempotent<int, Disposable>();
bool r = idempotent.TryCreateLifetime(1, k => new Disposable(), out var lifetime);
```

```cs
var idempotent = new ScopedAsyncIdempotent<int, Disposable>();
(bool r, Lifetime<Disposable> l) result = await idempotent.TryCreateLifetimeAsync(1, k =>
            {
                return Task.FromResult(new Disposable);
            });
```

If an exception is thrown, it is not cached. Intent is to mimic behavior of ConcurrentDictionary.GetOrAdd. In the case of the async variants, the result of the single failing task can be seen by other threads if they attempt to create a value while it is running. Once it is finished, a new attempt can be made.

This would probably need to be extended to include some parameterization of the lock timeout. This would require switching to Monitor.TryEnter.

### DIFF
--- a/BitFaster.Caching.UnitTests/Synchronized/AsyncIdempotentTests.cs
+++ b/BitFaster.Caching.UnitTests/Synchronized/AsyncIdempotentTests.cs
@@ -10,12 +10,12 @@ using Xunit;
 
 namespace BitFaster.Caching.UnitTests.Synchronized
 {
-    public class AsyncAtomicTests
+    public class AsyncIdempotentTests
     {
         [Fact]
         public void DefaultCtorValueIsNotCreated()
         {
-            var a = new AsyncAtomic<int, int>();
+            var a = new AsyncIdempotent<int, int>();
 
             a.IsValueCreated.Should().BeFalse();
             a.ValueIfCreated.Should().Be(0);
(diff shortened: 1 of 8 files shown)


BitFaster.Caching/Synchronized/Atomic.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace BitFaster.Caching.Synchronized
+{
+    [DebuggerDisplay("IsValueCreated={IsValueCreated}, Value={ValueIfCreated}")]
+    public class Atomic<K, V>
```
bitfaster, line 12: This is probably the wrong name - it's like indivisible smallest unit rather than exactly once.

Probably it should be called:

- Synchronized
- ScopedSynchronized
- AsyncSynchronized
- ScopedAsyncSynchronized

It's kind of weird having AsyncSynchronized etc. though.

PR #189, opened 2022-08-20
rawbuffer

bitfaster:
Provide API for both TryAdd/Take with spin loop or TryAddRaw/TakeRaw which exposes buffer status.

This makes it easier to write the background scheduler code, which should always retry on contention.


BitFaster.Caching/Buffers/BoundedBuffer.cs
```diff
@@ -148,7 +184,7 @@ public BufferStatus TryTake(out T item)
             return BufferStatus.Contended;
         }
 
-        public BufferStatus TryAdd(T item)
+        public BufferStatus TryAddRaw(T item)
```
bitfaster, line 187: TryAddFailFast?

PR #197, opened 2022-08-26
mpsc buffer

bitfaster:
A simplified multiple producer single consumer buffer, and striped version based on Caffeine BoundedBuffer. The drain to array method is more efficient. Latency is better for both background and foreground.

https://github.com/bitfaster/BitFaster.Caching/pull/197/commits/e34a2696b06e4c678c9ef787b75648e06e2e31ec - not sure how representative this run is - due to variability, it did _a lot_ of reps on ConcurrentLfuBackground on .NET6.
|                  Method |            Runtime |      Mean |     Error |    StdDev | Ratio | Allocated |
|------------------------ |------------------- |----------:|----------:|----------:|------:|----------:|
|    ConcurrentDictionary |           .NET 6.0 |  7.745 ns | 0.0335 ns | 0.0313 ns |  1.00 |         - |
| ConcurrentLfuBackground |           .NET 6.0 | 40.975 ns | 1.6332 ns | 4.7641 ns |  5.66 |         - |
|  ConcurrentLfuForeround |           .NET 6.0 | 67.327 ns | 0.4557 ns | 0.4262 ns |  8.69 |         - |
| ConcurrentLfuThreadPool |           .NET 6.0 | 70.637 ns | 0.6907 ns | 0.6461 ns |  9.12 |         - |
|       ConcurrentLfuNull |           .NET 6.0 | 26.625 ns | 0.1291 ns | 0.1144 ns |  3.44 |         - |
|                         |                    |           |           |           |       |           |
|    ConcurrentDictionary | .NET Framework 4.8 | 14.294 ns | 0.0911 ns | 0.0807 ns |  1.00 |         - |
| ConcurrentLfuBackground | .NET Framework 4.8 | 76.374 ns | 1.5344 ns | 1.7054 ns |  5.33 |         - |
|  ConcurrentLfuForeround | .NET Framework 4.8 | 79.986 ns | 0.8199 ns | 0.7669 ns |  5.59 |         - |
| ConcurrentLfuThreadPool | .NET Framework 4.8 | 76.536 ns | 0.8888 ns | 0.8314 ns |  5.35 |         - |
|       ConcurrentLfuNull | .NET Framework 4.8 | 41.019 ns | 1.8125 ns | 5.3441 ns |  3.11 |         - |

https://github.com/bitfaster/BitFaster.Caching/pull/197/commits/8515cd05fc4c24c4174fb87c9d5268be67702b6f - switch to volatile head and tail. Also many reps on ConcurrentLfuBackground - high error.

|                  Method |            Runtime |      Mean |     Error |    StdDev | Ratio | Allocated |
|------------------------ |------------------- |----------:|----------:|----------:|------:|----------:|
|    ConcurrentDictionary |           .NET 6.0 |  7.771 ns | 0.0446 ns | 0.0417 ns |  1.00 |         - |
| ConcurrentLfuBackground |           .NET 6.0 | 37.830 ns | 1.2401 ns | 3.6370 ns |  5.08 |         - |
|  ConcurrentLfuForeround |           .NET 6.0 | 67.874 ns | 0.1870 ns | 0.1562 ns |  8.73 |         - |
| ConcurrentLfuThreadPool |           .NET 6.0 | 52.346 ns | 1.0792 ns | 1.3253 ns |  6.74 |         - |
|       ConcurrentLfuNull |           .NET 6.0 | 25.323 ns | 0.0947 ns | 0.0840 ns |  3.26 |         - |
|                         |                    |           |           |           |       |           |
|    ConcurrentDictionary | .NET Framework 4.8 | 14.231 ns | 0.0805 ns | 0.0753 ns |  1.00 |         - |
| ConcurrentLfuBackground | .NET Framework 4.8 | 75.578 ns | 1.2353 ns | 1.1555 ns |  5.31 |         - |
|  ConcurrentLfuForeround | .NET Framework 4.8 | 79.652 ns | 0.2798 ns | 0.2617 ns |  5.60 |         - |
| ConcurrentLfuThreadPool | .NET Framework 4.8 | 58.151 ns | 0.7469 ns | 0.6621 ns |  4.08 |         - |
|       ConcurrentLfuNull | .NET Framework 4.8 | 36.992 ns | 0.2244 ns | 0.2099 ns |  2.60 |         - |

main

|                  Method |            Runtime |      Mean |     Error |     StdDev | Ratio | Allocated |
|------------------------ |------------------- |----------:|----------:|-----------:|------:|----------:|
|    ConcurrentDictionary |           .NET 6.0 |  7.530 ns | 0.0405 ns |  0.0359 ns |  1.00 |         - |
| ConcurrentLfuBackground |           .NET 6.0 | 51.408 ns | 3.5763 ns | 10.5449 ns |  8.26 |         - |
|  ConcurrentLfuForeround |           .NET 6.0 | 77.045 ns | 1.5122 ns |  1.9663 ns | 10.39 |         - |
| ConcurrentLfuThreadPool |           .NET 6.0 | 77.638 ns | 1.2778 ns |  1.1953 ns | 10.30 |         - |
|       ConcurrentLfuNull |           .NET 6.0 | 28.605 ns | 0.4603 ns |  0.4305 ns |  3.80 |         - |
|                         |                    |           |           |            |       |           |
|    ConcurrentDictionary | .NET Framework 4.8 | 16.090 ns | 0.0512 ns |  0.0427 ns |  1.00 |         - |
| ConcurrentLfuBackground | .NET Framework 4.8 | 78.216 ns | 1.0883 ns |  1.0180 ns |  4.86 |         - |
|  ConcurrentLfuForeround | .NET Framework 4.8 | 83.023 ns | 0.3906 ns |  0.3462 ns |  5.16 |         - |
| ConcurrentLfuThreadPool | .NET Framework 4.8 | 64.026 ns | 0.4500 ns |  0.3989 ns |  3.98 |         - |
|       ConcurrentLfuNull | .NET Framework 4.8 | 37.467 ns | 0.1786 ns |  0.1583 ns |  2.33 |         - |

Throughput is not obviously better one way or another (this is a very rudimentary test, new zipf each time influences results):

image

### DIFF
--- a/BitFaster.Caching/Buffers/MpscBoundedBuffer.cs
+++ b/BitFaster.Caching/Buffers/MpscBoundedBuffer.cs
@@ -1,0 +1,162 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace BitFaster.Caching.Buffers
+{
+    /// <summary>
+    /// A multiple producer single consumer buffer.
+    /// </summary>
+    public class MpscBoundedBuffer<T> where T : class
+    {
+        private T[] buffer;
+        private readonly int mask;
+        private PaddedHeadAndTail headAndTail; // mutable struct, don't mark readonly
+
+        public MpscBoundedBuffer(int boundedLength)
+        {
+            if (boundedLength < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(boundedLength));
+            }
+
+            // must be power of 2 to use & slotsMask instead of %
+            boundedLength = BitOps.CeilingPowerOfTwo(boundedLength);
+
+            buffer = new T[boundedLength];
+            mask = boundedLength - 1;
+        }
+
+        public int Capacity => buffer.Length;
+
+        public int Count
+        {
+            get
+            {
+                var spinner = new SpinWait();
+                while (true)
+                {
+                    var headNow = Volatile.Read(ref headAndTail.Head);
+                    var tailNow = Volatile.Read(ref headAndTail.Tail);
+
+                    if (headNow == Volatile.Read(ref headAndTail.Head) &&
+                        tailNow == Volatile.Read(ref headAndTail.Tail))
+                    {
+                        return GetCount(headNow, tailNow);
+                    }
+
+                    spinner.SpinOnce();
+                }
+            }
+        }
+
+        private int GetCount(int head, int tail)
+        {
+            if (head != tail)
+            {
+                head &= mask;
+                tail &= mask;
+
+                return head < tail ? tail - head : buffer.Length - head + tail;
+            }
+            return 0;
+        }
+
+        // thread safe
+        public BufferStatus TryAdd(T item)
+        {
+            int head = this.headAndTail.Head;
+            int tail = Volatile.Read(ref this.headAndTail.Tail);
+            int size = tail - head;
+
+            if (size >= buffer.Length)
+            {
+                return BufferStatus.Full;
+            }
+
+            if (Interlocked.CompareExchange(ref this.headAndTail.Tail, tail + 1, tail) == tail)
+            {
+                int index = (int)(tail & mask);
+                Volatile.Write(ref buffer[index], item);
+
+                return BufferStatus.Success;
+            }
+
+            return BufferStatus.Contended;
+        }
+
+        // thread safe for try take + multiple try add
+        public BufferStatus TryTake(out T item)
+        {
+            int head = this.headAndTail.Head;
+            int tail = Volatile.Read(ref this.headAndTail.Tail);
+            int size = tail - head;
+
+            if (size == 0)
+            {
+                item = default;
+                return BufferStatus.Empty;
+            }
+
+            int index = head & mask;
+
+            item = Volatile.Read(ref buffer[index]);
+
+            if (item == null)
+            {
+                // not published yet
+                return BufferStatus.Contended;
+            }
+
+            Volatile.Write(ref buffer[index], null);
+            Volatile.Write(ref this.headAndTail.Head, head + 1);
+            return BufferStatus.Success;
+        }
+
+        // thread safe for single drain + multiple try add
+        public int DrainTo(ArraySegment<T> output)
+        {
+            int head = this.headAndTail.Head;
+            int tail = Volatile.Read(ref this.headAndTail.Tail);
+            int size = tail - head;
+
+            if (size == 0)
+            {
+                return 0;
+            }
+
+            int outCount = 0;
+
+            do
+            {
+                int index = head & mask;
+
+                T item = Volatile.Read(ref buffer[index]);
+
+                if (item == null)
+                {
+                    // not published yet
+                    break;
+                }
+
+                Volatile.Write(ref buffer[index], null);
+                output.Array[output.Offset + outCount++] = item;
+                head++;
+            }
+            while (head != tail && outCount < output.Count);
+
+            Volatile.Write(ref this.headAndTail.Head, head);
+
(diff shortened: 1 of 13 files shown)


BitFaster.Caching/Buffers/MpscBoundedBuffer.cs
```diff
@@ -0,0 +1,167 @@
...
+                return head < tail ? tail - head : buffer.Length - head + tail;
+            }
+            return 0;
+        }
+
+        // thread safe
+        public BufferStatus TryAdd(T item)
+        {
+            int head = this.headAndTail.Head;
+            int tail = Volatile.Read(ref this.headAndTail.Tail);
```
ben-manes, line 76: fwiw, I did a volatile read for a happens-before edge (acquire / load-load) so that the other plain read could piggyback on it for free. You have it before so you are more likely to get a stale read. Probably doesn’t matter much though.

bitfaster: Piggybacking is some next level thinking.

I hadn't spotted that readCounter is declared volatile in your code (which was careless because I thought I was paying attention). 

I don't know if .NET has the same visibility guarantee as Java - whether non volatile fields are also refreshed. I need to think hard about this. The Volatile.Read static method actually inserts a full fence afaik, so probably VolatileRead(head) first would do the trick here.

bitfaster: I switched both head and tail to volatile. So now there are 2 acquire reads, instead of 1 full fence and 1 plain (since I no longer call Volatile.Read or Volatile.Write which each internally call Thread.MemoryBarrier). Eliminating the full fences makes it about 7.5% faster in the microbenchmark. 

In .NET there is no built-in way to do a non-volatile read of a volatile field, or a volatile read of a non-volatile field without inserting a full fence. The APIs are not as good as Java or C++ which are much more explicit and give all options.

The only workaround I can think of would be to make a union by placing 2 int fields in the PaddedHeadAndTail struct with the same offset and make only 1 volatile. I feel like that could work, but this is all pretty esoteric stuff so I'm not certain.

The other problem is that head and tail are integers, and I think this code will break when they wrap.

ben-manes: I think it is probably fine as volatile reads. I suppose technically one could claim the map lookup provided an early enough acquire fence, as this is called only on a hit. I play with memory barriers to learn the hardware, but I don't think it measurably impacts performance. My comment was more about possible confusion by swapping the order.

I don't recall why I used a `long`, maybe out of habit? It shouldn't break due to an overflow since it uses the delta as a guard and a mask for the index. It's a good test case that I should to add, though.

bitfaster: You're right. I thought the size computation would break, but actually I think it is correct because when tail has wrapped to -2147483647 subtracting a large head number will itself wrap the result back (overflow is unchecked by default).

ben-manes: fwiw, I borrowed the striping logic for buffers from Java's concurrent counters (LongAdder). Those counters are what I use if metrics are enabled, e.g. the hit counter. Since the read buffer is lossy it didn't seem appropriate to rely on maintenance. The lazily initialized, scalable counters serve the same purpose of trying to avoid memory waste and trying to give each cpu its own counter to minimize CAS conflicts. This of course has to retry until incremented and the counters summed for the total value.

I raise this because if you end up writing a similarly striped class for the read buffers, you are 90% there for scalable counters. While enabling stats will degrade somewhat, it won't be nearly as substantial so as to make a difference for end users. That's a useful utility class in its own right, so I'm surprised .NET doesn't have one already and your own could be a useful standalone utility.

PR #204, opened 2022-08-27
nostripew

bitfaster:
1. Don't use a striped buffer for writes
2. Cap the write buffer size to nextPowerOf2(cache size), or 128. Whichever is smaller.

This is mostly similar throughput, a little better when thread count is lower.
image


BitFaster.Caching.UnitTests/Lfu/ConcurrentLfuTests.cs
```diff
@@ -421,18 +415,18 @@ public void WhenWriteBufferIsFullAddDoesMaintenance()
 
             // add an item, flush write buffer
             cache.GetOrAdd(-1, k => k);
-            scheduler.RunCount.Should().Be(1);
```
bitfaster, line 424: This is probably wrong.

PR #246, opened 2022-09-09
unsafe

bitfaster:
Try to eliminate bounds checks with unsafe code.

Both unsafe variants are approximately equal throughput in Read+Write compared to using the standard array.


bitfaster:
These bounds checks are already eliminated, per
https://learn.microsoft.com/en-us/archive/blogs/clrcodegeneration/array-bounds-check-elimination-in-the-clr

PR #265, opened 2022-09-17
bounds

bitfaster:
Remove bounds checks by capturing a local ref and iterating backwards.

First read about this here:
https://www.meziantou.net/some-performance-tricks-with-dotnet-strings.htm


bitfaster:
These are eliminated, documented here:
https://learn.microsoft.com/en-us/archive/blogs/clrcodegeneration/array-bounds-check-elimination-in-the-clr

PR #280, opened 2022-09-29
faster zipf

bitfaster:
This is about 20x faster than using Math.Numerics, since the harmonic is computed once and the loop that generates the values is now parallelized. 

Only the random generation is serial - this means that the output is identical Math.Numerics since the same 'random' numbers are generated.


BitFaster.Caching.HitRateAnalysis/BitFaster.Caching.HitRateAnalysis.csproj
```diff
@@ -26,4 +30,8 @@
     <ProjectReference Include="..\BitFaster.Caching\BitFaster.Caching.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Properties\" />
+  </ItemGroup>
```
bitfaster, line 35: remove

PR #292, opened 2022-10-04
multi-target tests

bitfaster:
Cover all target frameworks on each platform.

Problems:

- Need to upload code coverage separately for each target framework
- Test checks are now based on several .trx files where the file name is based on a timestamp, so can't be hardcoded
- .NET48 generates 0% coverage

This could be an option for multi-trx:
https://github.com/im-open/process-dotnet-test-results


bitfaster:
Superseded by https://github.com/bitfaster/BitFaster.Caching/pull/342

PR #342, opened 2023-04-21
multi-target unit tests

bitfaster:
This PR:

1. Enables unit tests to build and run on .NET 4.8, .NET Core 3.1 and .NET 6 targets.
2. Runs all targets on windows
3. Runs .NET6 on mac
4. Collects code coverage for 3.1 and 6.0, since .NET Framework is not well supported in dotnet test.


bitfaster:
Looks like coverage result files will have target fwk like this:

image

PR #343, opened 2023-04-22
validate package

bitfaster:
See https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview

Diagnostic IDs are here: https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/diagnostic-ids


bitfaster:
Fails with:

<html>
<body>


CP0007 | A base type on the class hierarchy was removed from one of the compared sides. | Add the base type back. (A new base type can be introduced in the hierarchy if that's intended.)
-- | -- | --



</body>
</html>

PR #370, opened 2023-07-18
unbounded

bitfaster:
Fix unbounded LRU growth when keys are requested in order. Added better parameterized tests to detect this.

I had introduced a bug in the cycle/tryremovecold method: when warm queue is full the last cold item should be discarded immediately.

When cache size is below 1000, element count is now constrained to capacity +1 (we allow warm queue to be capacity+1, enabling items to re-enter warm). When cache size is above 1000, element count is constrained to at worst capacity + 5. Experimentally, this matches the number of cycle attempts (which is currently 5). This is a bug that results in at most a 0.5% capacity overage, so much better than prior to the fix.


bitfaster:
Verified hit rate analysis is approx the same (difference is noise), except for ARC OLTP which is around 1% better hit rate across the board.

PR #378, opened 2023-08-17
remove coldest when full

bitfaster:
Fix cycle logic so that if all items are touched when a new item is added (thus the coldest is cycled back into warm) we simply discard the coldest cache item (last item in cold queue).


bitfaster:
Hit rate tests

## Glimpse

| CacheSize | Before | After |
|-----------|--------------------|----------------------|
|       250 |              12.83 |   12.834580216126351 |
|       500 |              26.13 |   26.134663341645886 |
|       750 |              39.43 |    39.43474646716542 |
|      1000 |              43.28 |   43.275145469659186 |
|      1250 |              49.93 |   49.925187032418954 |
|      1500 |              52.4  |    52.40232751454697 |
|      1750 |              53.5  |    53.49958437240233 |
|      2000 |              53.55 |    53.54945968412302 |

bitfaster:
With cold/warm reversed in the cycle method, the throughput test (multi-threaded) results in some runaway growth. Restoring the original sequence fixes this. Throughput is marginally lower - need to test on a clean machine without other processes running.

PR #385, opened 2023-08-22
rem braces

bitfaster:
Removing braces around `if(...) { throw; }` to eliminate the uncovered code:
image
image

There is a change to coverlet that states decorating the throw helper methods with `[DoesNotReturn]` should eliminate this (imlpemented in https://github.com/bitfaster/BitFaster.Caching/pull/386), but it doesn't work. Therefore, simply remove the braces to avoid artificially counting code coverage lower.


bitfaster:
With change:

image
image

PR #388, opened 2023-08-23
cover switch expression

bitfaster:
Slightly confused about how to cover this switch expression, the tests appear to cover all branches:

image

This is the lowered code according to dotPeek:

image


bitfaster:
See this for discussion on fixed:
https://reddit.com/r/csharp/s/kODlobHKvH

bitfaster:
This is now showing as covered:
image

PR #389, opened 2023-08-24
soak singletoncache

bitfaster:
Attempt to cover this code:

image


bitfaster:
This is not reachable.

PR #401, opened 2023-10-02
ConcurrentLfu.Clear() polluted by removed items

bitfaster:
Calling `Clear()` with `BackgroundThreadScheduler` can result in a situation where window/probation/protected contains an item that was already removed. Since the clear logic trims n items, where n == cache.Count, the already removed items pollute the list of candidates resulting in non-removed items remaining in the cache.

### DIFF
--- a/BitFaster.Caching.UnitTests/Lfu/ConcurrentLfuTests.cs
+++ b/BitFaster.Caching.UnitTests/Lfu/ConcurrentLfuTests.cs
@@ -677,64 +677,34 @@ namespace BitFaster.Caching.UnitTests.Lfu
         [Fact]
         public void WhenClearedCacheIsEmpty()
         {
-            int key = 0;
-            const int count = 20;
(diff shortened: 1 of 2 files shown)


BitFaster.Caching.UnitTests/Lfu/ConcurrentLfuTests.cs
```diff
@@ -677,15 +677,64 @@ public void WhenItemDoesNotExistTryUpdateIsFalse()
...
-            cache.TryGet(1, out var _).Should().BeFalse();
+                cache.Clear();
+
+                cache.Count.Should().Be(0);
+                cache.TryGet(1, out var _).Should().BeFalse();
+            }
+        }
+
+        [Fact]
+        public async Task TestClear()
```
bitfaster, line 702: Change this to non-async + non-atomic and verify repro.

PR #409, opened 2023-10-14
Scoped cache broken due to operator== ABA problem

bitfaster:
See the [wikipedia description for the ABA problem](https://en.wikipedia.org/wiki/ABA_problem).

At some point, ReferenceCount was updated to implement IEquatable (possibly because code analysis said this is a good idea). This changed `operator ==` from the default which tests whether two references of that type refer to the same object, to comparing for field equality. 

This broke the `Scoped` class. Specifically, [this line](https://github.com/bitfaster/BitFaster.Caching/blob/5ffc05a6317ad1af4874d59ad8a45c381b0c0f02/BitFaster.Caching/Scoped.cs#L54) and [this line](https://github.com/bitfaster/BitFaster.Caching/blob/5ffc05a6317ad1af4874d59ad8a45c381b0c0f02/BitFaster.Caching/Scoped.cs#L83) are now susceptible to the ABA problem because the equality operator is checking field values:

`if (oldRefCount == Interlocked.CompareExchange(ref this.refCount, oldRefCount.IncrementCopy(), oldRefCount))`

This issue does not affect SingletonCache, because it does not depend on `operator==` or `operator!=`.

There was a further bug in `Scoped.DecrementReferenceCount()` caused by reading a stale value of `this.refCount.Count` after it was replaced by a compare exchange call.

### DIFF
--- a/BitFaster.Caching/Scoped.cs
+++ b/BitFaster.Caching/Scoped.cs
@@ -82,6 +82,8 @@ namespace BitFaster.Caching
 
                 if (oldRefCount == Interlocked.CompareExchange(ref this.refCount, oldRefCount.DecrementCopy(), oldRefCount))
                 {
+                    // Note this.refCount may be stale.
+                    // Old count == 1, thus new ref count is 0, dispose the value.
                     if (oldRefCount.Count == 1)
                     {
                         oldRefCount.Value?.Dispose();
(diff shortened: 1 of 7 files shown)


BitFaster.Caching/Scoped.cs
```diff
@@ -45,7 +45,7 @@ public bool TryCreateLifetime(out Lifetime<T> lifetime)
                 var oldRefCount = this.refCount;
 
                 // If old ref count is 0, the scoped object has been disposed and there was a race.
-                if (this.isDisposed || oldRefCount.Count == 0)
+                if (IsDisposed || oldRefCount.Count == 0)
```
bitfaster, line 48: I think there is another potential race as follows:

1. Thread A: check IsDisposed returns false
2. Thread B: Dispose, IsDisposed now true
3. Thread A: IncrementCount, return lifetime containing disposed value

I am not able to provoke this case with the soak test.

PR #411, opened 2023-10-14
Extend timeout for mpsc buffer test

bitfaster:
It seems like the core problem is that xunit will start to heavily parallelize test execution, and the github action runs on a 2 core machine. Previous bug fix PR added two more soak tests, which led to more severe thread starvation. Increasing the timeout makes this much more stable. Next step will be to break out soak tests and [not run them in parallel](https://xunit.net/docs/running-tests-in-parallel).

- Removed use of `[Fact(Timeout=x]` because it is documented as undefined behavior when tests are run in parallel
- Added `Task.TimeoutAfter` extension method. This is likely not better, because the real issue is xunit scheduling too many threads.
- Cleaned up using statement/redundant cast in `MpscBoundedBuffer`

### DIFF
--- a/BitFaster.Caching.UnitTests/Buffers/MpscBoundedBufferTests.cs
+++ b/BitFaster.Caching.UnitTests/Buffers/MpscBoundedBufferTests.cs
@@ -13,7 +13,7 @@ namespace BitFaster.Caching.UnitTests.Buffers
 {
     public class MpscBoundedBufferTests
     {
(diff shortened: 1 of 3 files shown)


bitfaster:
It could be that this is caused by the timeout attribute:

image

bitfaster:
TimeoutAfter taken from [here](https://devblogs.microsoft.com/pfxteam/crafting-a-task-timeoutafter-method/).

PR #416, opened 2023-10-14
guard against delays during tests that depend on wall clock time

bitfaster:
In the GitHub action, we see failures like this:

```
[xUnit.net 00:00:12.35]     ConcurrentTLruDefaultClockTests.WhenCacheHasExpiredAndFreshItemsExpireRemovesOnlyExpiredItems [FAIL]
  Failed ConcurrentTLruDefaultClockTests.WhenCacheHasExpiredAndFreshItemsExpireRemovesOnlyExpiredItems [379 ms]
  Error Message:
   Expected lru.Count to be 3, but found 2.
```

This shows the test ran for 379 ms. But the TTL for deleting items is 10ms wall clock time. If the test experiences a delay like this (which is unlikely), extra items can expire and the test will fail.

Instead, check the end to end wall clock time before doing the assert. If there was a delay, try again (up to 64 times). Given that this failure is very seldom, this should result in very good stability.

It's arguably better to inject a fake clock, but the reason this test exists was due to expiry being broken only on MacOS while tests with a fake clock passed.


bitfaster:
Superseded by https://github.com/bitfaster/BitFaster.Caching/pull/468

PR #422, opened 2023-10-16
better initial size

bitfaster:
To minimize hash collisions, the internal `ConcurrentDictionary` bucket count must be:

- A prime number.
- 33% greater than the total number of items in the cache (or have a [load factor of 0.75](https://en.wikipedia.org/wiki/Hash_table)).

`ConcurrentDictionary` does a less than ideal job of choosing HashBucket count until .NET8 (when it was updated to match `Dictionary` and use prime numbers). So, on earlier versions of the framework we want to pass a value to the `ConcurrentDictionary` ctor that will minimize collisions.

PR https://github.com/bitfaster/BitFaster.Caching/pull/421 introduced a simple scheme to improve bucket count, but this can still waste a lot of memory for larger capacity caches due to `ConcurrentDictionary` 2x + 1 resize scheme, and likely won't grow to a bucket count near the ideal 133% of the cache size.

To solve this, introduce an Estimate method that inflates the capacity by 33%, then uses a hybrid scheme to select the initial size such that:

- For small caches, start with a `ConcurrentDictionary` capacity that is a prime number 33% larger than the number of cache items. Initial size is large enough to avoid resize.
- For large caches, pick an initial size using a lookup table. Initial size is approximately 10% of the desired size such that 4 `ConcurrentDictionary` resize operations will arrive at a hash table size that is a prime number approximately 33% larger than cache capacity.

This PR includes the source code for the tool that was used to generate the lookup table.


BitFaster.Caching/HashTableSize.cs
```diff
@@ -0,0 +1,656 @@
...
+#endif
+            7, 11, 17, 23, 29, 37, 47, 59, 71, 89, 107, 131, 163, 197
+        };
+
+#if NETSTANDARD2_0
+        internal static KeyValuePair<int, int>[] SizeMap =
+#else
+        internal static ReadOnlySpan<KeyValuePair<int, int>> SizeMap =>
+#endif
+            new KeyValuePair<int, int>[576]
```
bitfaster, line 76: This is about 4.6kb - probably need to shrink it to say 1kb.

bitfaster: 129 ints is about 1.032 kb.

BitFaster.Caching/HashTableSize.cs
```diff
@@ -0,0 +1,209 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace BitFaster.Caching
+{
+    // Using the capacity passed into the cache ctor to initialize the ConcurrentDictionary has 2 problems:
```
bitfaster, line 6: Update this - it reflects old implementation.

PR #425, opened 2023-10-17
singleton cache capacity is a prime number

bitfaster:
Use prime number lookup to initialize `SingletonCache` capacity.

The lookup table now replicates the internal .NET framework [HashHelpers](https://source.dot.net/System.Collections.Concurrent/R/de3ba4873d4ad06a.html), which caps out at 7,199,369. Above this size, the raw input argument is used. Setting a prime number for hash table bucket count is best efforts: this PR optimizes the common case, relatively small `SingletonCache`. Avoid increasing the lookup table size further or adding code to compute primes.


BitFaster.Caching/ConcurrentDictionarySize.cs
```diff
@@ -15,7 +15,7 @@ internal static int NextPrimeGreaterThan(int min)
                 }
             }
 
-            return 197;
+            return 7199369;
```
bitfaster, line 18: We could simply return the input here, which is equivalent to the existing behavior. That way this is best efforts.

PR #444, opened 2023-10-25
Enable AOT trimming

bitfaster:
Enable trimming compatibility:
https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming


bitfaster:
Looks like this is not needed, no trim warmings.

PR #448, opened 2023-11-02
remove Lfu half fence reads/writes

bitfaster:
Successive buffer reads/writes don't need two volatile ops - they can piggyback. This was originally suggested by Ben but wasn't stable at the time. Now able to verify stability via soak tests.

TODO:
1. Bench
2. Throughput


bitfaster:
This change seems to be consistently better for evict, but worse for r/w ([c03602d](https://github.com/bitfaster/BitFaster.Caching/pull/448/commits/c03602d81ba5176c61b9918dbdf0d60c407a9609)):

image
image

bitfaster:
image
image

PR #467, opened 2023-11-12
ConcurrentLru item based time eviction policy

bitfaster:
Implement a time-based eviction policy where the caller can specify the time to expire per item. The `IExpiryCalculator` interface enables the caller to evaluate both the key and the value to compute a `TimeSpan` that defines the time to expire:

```cs
public class Expiry : IExpiryCalculator<string, int>
{
    public TimeSpan GetExpireAfterCreate(string key, int value) => TimeSpan.FromSeconds(10);
    public TimeSpan GetExpireAfterRead(string key, int value, TimeSpan current) => current;
    public TimeSpan GetExpireAfterUpdate(string key, int value, TimeSpan current) => current;
}

ICache<string, int> expireAfter = new ConcurrentLruBuilder<string, int>()
   .WithExpireAfter(new Expiry())
   .Build();
```

This is the same API used by [Caffeine](https://github.com/ben-manes/caffeine/wiki/Eviction).

The new policy is exposed on `ICache` via `CachePolicy` as a new `Optional<IDiscreteTimePolicy>` called `ExpireAfter`:

```cs
public class CachePolicy
{
    // ...
    public Optional<IBoundedPolicy> Eviction { get; }
    public Optional<ITimePolicy> ExpireAfterWrite { get; }
    public Optional<ITimePolicy> ExpireAfterAccess { get; }
    public Optional<IDiscreteTimePolicy> ExpireAfter { get; }
}
```

where `IDiscreteTimePolicy` is defined as

```cs
    public interface IDiscreteTimePolicy
    {
        bool TryGetTimeToExpire<K>(K key, out TimeSpan timeToExpire);
        void TrimExpired();
    }
```

- There are two places where it would be a breaking change to get end to end generics:
   1. `LruInfo` does not know about value type, but `IExpiry` does. Therefore, internally store as object and use generic get and set methods to specify the expiry. The builder methods are typed, so this is largely hidden behind the builder.
   2. `IVariableTimePolicy.TryGetTimeToLive()` has key type as a generic _method_ parameter, because Policy is not generic in terms of key or value.
- `WithExpireAfter` is present only on `ConcurrentLruBuilder` and `AsyncConcurrentLruBuilder`
   - You can't call `WithExpireAfter` on `AtomicAsyncConcurrentLruBuilder`, `AtomicConcurrentLruBuilder`, `AtomicScopedAsyncConcurrentLruBuilder`, `AtomicScopedConcurrentLruBuilder`, `ScopedAsyncConcurrentLruBuilder` or `ScopedConcurrentLruBuilder`.
   - It is still possible to call `WithExpireAfter`, then call an As method to change the builder type. The above builders all have guards in the build method to throw a runtime exception in this case.


BitFaster.Caching.UnitTests/Lru/ConcurrentLruBuilderTests.cs
```diff
@@ -178,6 +178,20 @@ public void TestExpireAfterReadAndExpireAfterWriteThrows()
             act.Should().Throw<InvalidOperationException>();
         }
 
+        [Fact]
+        public void TestExpireAfter()
+        {
+            ICache<string, int> expireAfter = new ConcurrentLruBuilder<string, int>()
+                .WithExpireAfter(new Expiry<string, int>((k, v) => TimeSpan.FromMinutes(5)))
```
bitfaster, line 185: Should this simply be `WithExpiry`?

BitFaster.Caching/CachePolicy.cs
```diff
@@ -47,6 +62,35 @@ public CachePolicy(Optional<IBoundedPolicy> eviction, Optional<ITimePolicy> expi
...
-        public Optional<ITimePolicy> ExpireAfterAccess { get; }
+        public Optional<ITimePolicy> ExpireAfterAccess { get; }
+
+        /// <summary>
+        /// Gets the expire after policy, if any. This policy evicts items after a
+        /// variable time to live computed from the key and value.
+        /// </summary>
+        public Optional<IVariableTimePolicy> ExpireAfter { get; }
+
+        public bool TryTrimExpired()
```
bitfaster, line 73: This should be done independently.

BitFaster.Caching/Lru/Builder/LruInfo.cs
```diff
@@ -28,7 +28,12 @@ public sealed class LruInfo<K>
         /// <summary>
         /// Gets or sets the time to expire after access.
         /// </summary>
-        public TimeSpan? TimeToExpireAfterAccess { get; set; } = null;
+        public TimeSpan? TimeToExpireAfterAccess { get; set; } = null;
+
+        /// <summary>
+        /// Gets or sets the custom expire after.
```
bitfaster, line 34: The Expiry?

BitFaster.Caching/Lru/CustomPolicy.cs
```diff
@@ -0,0 +1,293 @@
...
+    {
+        /// <summary>
+        /// Gets the expiry for an item after it is created.
+        /// </summary>
+        Func<K, V, TimeSpan> GetExpireAfterCreate { get; }
+
+        /// <summary>
+        /// Gets the expiry for an item after it is read.
+        /// </summary>
+        Func<K, V, TimeSpan> GetExpireAfterRead { get; }
```
bitfaster, line 20: This cannot accept the existing TTL as input, because we have converted it to ticks since the machine started which is non-reversable.

BitFaster.Caching/ITimePolicy.cs
```diff
@@ -1,24 +1,40 @@
...
+        /// Remove all expired items from the cache.
+        /// </summary>
+        void TrimExpired();
+    }
+
+    /// <summary>
+    /// Represents a variable time based cache policy.
+    /// </summary>
+    /// backcompat: this should be generic in terms of Key to avoid object arg.
+    public interface IVariableTimePolicy
```
bitfaster, line 25: Alternatives:
`IItemTimePolicy`

BitFaster.Caching/Lru/CustomPolicy.cs
```diff
@@ -0,0 +1,293 @@
...
+{
+    /// <summary>
+    /// Defines a policy for determining the expiry of an item.
+    /// </summary>
+    public interface IExpiry<K, V>
+    {
+        /// <summary>
+        /// Gets the expiry for an item after it is created.
+        /// </summary>
+        Func<K, V, TimeSpan> GetExpireAfterCreate { get; }
```
bitfaster, line 15: This part shouldn't actually be a func, it should be a method on the interface.

That way, it is possible to implement this without delegates.

BitFaster.Caching/Lru/CustomPolicy.cs
```diff
@@ -0,0 +1,304 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+
+namespace BitFaster.Caching.Lru
+{
+    /// <summary>
+    /// Defines a mechanism to calculate when cache entries expire.
+    /// </summary>
+    public interface IExpiry<K, V>
```
bitfaster, line 9: `IExpiryCalculator`?

Consequently, ExpiryCalculator below.

BitFaster.Caching/Lru/CustomPolicy.cs
```diff
@@ -0,0 +1,305 @@
...
+        ///<inheritdoc/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public TimeSpan GetExpireAfterUpdate(K key, V value)
+        {
+            return this.expireAfterUpdate(key, value);
+        }
+    }
+
+#if NETCOREAPP3_0_OR_GREATER
+    internal readonly struct CustomExpiryPolicy<K, V> : IItemPolicy<K, V, LongTickCountLruItem<K, V>>
```
bitfaster, line 84: There is currently no way for someone to write an equivalent because ConcurrentLruCore binds to this internal type.

Ideally, there should be a marker interface that is public.

BitFaster.Caching/Lru/CustomPolicy.cs
```diff
@@ -0,0 +1,313 @@
...
+        ///<inheritdoc/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public TimeSpan GetExpireAfterUpdate(K key, V value, TimeSpan currentTtl)
+        {
+            return this.expireAfterUpdate(key, value);
+        }
+    }
+
+#if NETCOREAPP3_0_OR_GREATER
+    internal readonly struct DiscreteExpiryPolicy<K, V> : IItemPolicy<K, V, LongTickCountLruItem<K, V>>
```
bitfaster, line 86: item policy, not expiry policy

BitFaster.Caching/Lru/ConcurrentLruCore.cs
```diff
@@ -872,5 +877,35 @@ public void TrimExpired()
...
+                lru.TrimExpired();
+            }
+
+            public bool TryGetTimeToLive<TKey>(TKey key, out TimeSpan timeToLive)
+            {
+                if (key is K k && lru.dictionary.TryGetValue(k, out var item))
+                {
+                    if (item is LongTickCountLruItem<K, V> tickItem)
+                    {
+                        timeToLive = TimeSpan.FromTicks(tickItem.TickCount);
```
bitfaster, line 901: This is not correct on .NET Standard, which will use stopwatch ticks.

BitFaster.Caching/Lru/CustomPolicy.cs
```diff
@@ -0,0 +1,319 @@
...
+        /// Specify the time to live after an entry is updated.The current TTL may be
+        /// be returned to not modify the expiration time.
+        /// </summary>
+        TimeSpan GetExpireAfterUpdate(K key, V value, TimeSpan currentTtl);
+    }
+
+    /// <summary>
+    /// Defines a mechanism to determine the time to live for a cache item using function delegates.
+    /// </summary>
+    public readonly struct ExpiryCalculator<K, V> : IExpiryCalculator<K, V>
```
bitfaster, line 33: Need to figure out how to represent this cleanly so that:

1. You can give a single delegate (ignoring current TTL)
2. You can give all 3 delegates

Easiest way is to make this a base class, but perf will be bad. Fastest way is 2 structs, but what should they be called?

`SimpleExpiryCalculator` and `ExpiryCalculator`?

BitFaster.Caching/Lru/ConcurrentLruBuilder.cs
```diff
@@ -35,13 +35,32 @@ public ConcurrentLruBuilder()
...
         {
+        }
+
+        /// <summary>
+        /// Evict after a variable duration specified by an IExpiry instance.
+        /// </summary>
+        /// <param name="expiry">The expiry that determines item time to live.</param>
+        /// <returns>A ConcurrentLruBuilder</returns>
+        public ConcurrentLruBuilder<K, V> WithExpiry(IExpiryCalculator<K, V> expiry)
+        {
```
bitfaster, line 46: Should this be just an action delegate to start out:

`WithExpiry((k, v) => TimeSpan)`

bitfaster: Only interface is public, the delegate version is now only part of unit tests.

BitFaster.Caching/Lru/CustomPolicy.cs
bitfaster: Removed delegate based impl from the API surface.

BitFaster.Caching.UnitTests/Lru/DiscretePolicyTests.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        private static readonly ulong stopwatchDelta = (ulong)StopwatchTickConverter.ToTicks(TimeSpan.FromMilliseconds(20));
+        private static readonly ulong tickCountDelta = (ulong)TimeSpan.FromMilliseconds(20).ToEnvTick64();
+
+        private static readonly TimeSpan defaultTimeToExpire = TimeSpan.FromSeconds(10);
+
+        public DiscretePolicyTests() 
+        {
+            expiryCalculator = new TestExpiryCalculator<int, int>();
+            expiryCalculator.ExpireAfterCreate = (_, _) => defaultTimeToExpire;
```
bitfaster, line 23: This should be the default for the test calculator, so its not repeated. Default time to expire can be there as well.

PR #479, opened 2023-11-14
Provide ConcurrentLru factory

bitfaster:
This consolidates the duplicate creation logic between `AsyncConcurrentLruBuilder` and `ConcurrentLruBuilder`, and provides a smaller code size for AOT compilation (builder classes have a huge footprint).

What should this be called? Is it bad to have Create and CreateAsync? Make internal for now.


BitFaster.Caching/Lru/LruFactory.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System.Linq.Expressions;
+using BitFaster.Caching.Lru.Builder;
+
+namespace BitFaster.Caching.Lru
+{
+    /// <summary>
+    /// Factory class for creating ConcurrentLru variants.
+    /// </summary>
+    public static class LruFactory<K, V>
```
bitfaster, line 9: internal

PR #496, opened 2023-11-18
Optimize maintenance family of methods

bitfaster:
Try applying aggressive optimization to the maintenance methods. This is a tiny bit better for LFU, appears to be noise for LRU.

This would likely need to be removed if building on NET7 or NET8, because it can cancel PGO.

image

image


bitfaster:
Superseded by https://github.com/bitfaster/BitFaster.Caching/pull/530

PR #518, opened 2023-11-28
Use benchly

bitfaster:
Hookup benchly to automate generating Benchmark test results.

# .NET6 latency:

BitFaster Caching Benchmarks LruJustGetOrAdd- NET 6 0-columnchart

# Box plot for LFU:

BitFaster Caching Benchmarks LfuJustGetOrAdd-boxplot

### DIFF
--- a/BitFaster.Caching.Benchmarks/BitFaster.Caching.Benchmarks.csproj
+++ b/BitFaster.Caching.Benchmarks/BitFaster.Caching.Benchmarks.csproj
@@ -19,6 +19,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Benchly" Version="0.6.0" />
     <PackageReference Include="BenchmarkDotNet" Version="0.13.10" />
     <PackageReference Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.13.10" />
     <PackageReference Include="MathNet.Numerics" Version="5.0.0" />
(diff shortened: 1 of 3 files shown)


ben-manes:
I got a new laptop last week and this is at 16 threads 😁

image

bitfaster:
Wow - super impressive! Is it an M3 Max?

I am overdue a round of tests on .NET8 with dynamic PGO, but I doubt I will get close to that. I need a new laptop...

ben-manes:
haha, yes; the 14-core version. I hadn't upgraded my laptop since 2016 and the battery gave out so it became hard coerce it to power up. I didn't expect it to be that good!

bitfaster:
For reference I can only just hit those numbers on a monster desktop with an Intel 13900K, with > 24 threads (I think I showed you this last year):

image

bitfaster:
Thats an extremely impressive result for a 14 cores/threads - in addition to the hw the Java runtime/JIT must be well sorted.

A few weeks ago I got an M2 Mac mini to test ARM - I just did a quick check, with .NET8 ConcurrentLfu hits about 80 million/sec for me. That's a pretty substantial gap. The M3 max has 12 perf cores vs the M2's 4, but I'm still more than 10x slower.

ben-manes:
hmm.. that's interesting. I figured that since this is a cpu-bound workload where everything is in cache, the wider instruction decode engine might be giving it a large boost. On my 2016 laptop it was around 100M/s at 8 threads, iirc. I don't really think anything beyond a synthetic 60M/s is an actually visible benefit in the real world.

Maybe our benchmarks differ significantly (here is [mine](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/GetPutBenchmark.java))? I use a zipfian distribution to simulate a realistic request distribution with hot entries. This should take care of JIT warmup, by returning the value it should avoid dead code elimination (which would skew if untrue), and I tried to eliminate any benchmark overhead. What about yours?

bitfaster:
BenchmarkDotNet is the closest equivalent to Java Microbenchmark Harness, but it doesn't support threaded tests. Your tests are very clean - its nice you can specify threaded tests like that.

PR #519, opened 2023-11-30
Throughput analysis v2.0

bitfaster:
- Use BenchmarkDotNet's method to keep return values alive, mitigating dead code elimination for the return value from cache GetOrAdd. This reduces peak throughput by about 10%, likely due to the extra non-inlined method.
- Apply optimization option to test execution method.
- Use Threads instead of Tasks. This is more stable, but MemoryCache results now have more extreme outliers due to unpredictable nature of background thread eviction.
- Replicate the BenchmarkDotNet main loop to determine when the result is stable, based on computing statistics from the results so far.
- Use Perfolizer to eliminate outliers from the result calculation - this eliminates some of the outliers from MemoryCache when the eviction thread is not scheduled. The update and evict scenarios provoke unbounded growth/failure to evict for MemoryCache due to threadpool backlog.
- Adapt the number of iterations in each run so that a run is approximately 200ms. This gives a worst-case scenario time of about 35 secs, so an entire test run can complete on a 96 core machine in less than 1 day.
- Wait for the thread pool to empty before each run. This flushes out any backlog that MemoryCache has built up, making results more consistent.

---

Results for adaptive mode:

| Min iters | Max iters | Max Relative Error | Target Iteration time |
|-----------|-----------|----------------------|-----------------------|
| 15   | 80  | 0.02 | 200ms |

Results_Read_500

Results_ReadWrite_500

Results_Update_500

Results_Evict_500

### DIFF
--- a/BitFaster.Caching.ThroughputAnalysis/BitFaster.Caching.ThroughputAnalysis.csproj
+++ b/BitFaster.Caching.ThroughputAnalysis/BitFaster.Caching.ThroughputAnalysis.csproj
@@ -23,7 +23,7 @@
         <NoWarn>NU1701</NoWarn>
     </PackageReference>
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="8.0.0" />
-    <PackageReference Include="Plotly.NET.CSharp" Version="0.11.1" />
+    <PackageReference Include="Plotly.NET" Version="4.2.0" />
     <PackageReference Include="Plotly.NET.ImageExport" Version="5.0.1" />
   </ItemGroup>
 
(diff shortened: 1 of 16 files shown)


bitfaster:
---

Results for:

| Min iters | Max iters | Max Relative Error |
|-----------|-----------|----------------------|
| 6   | 25  | 0.04 | 

Results_Read_500
Results_ReadWrite_500
Results_Update_500
Results_Evict_500

---

Results for BenchmarkDotNet settings (now significantly slower for eviction test):

| Min iters | Max iters | Max Relative Error |
|-----------|-----------|----------------------|
| 15   | 80  | 0.02 | 

Results_Read_500
Results_ReadWrite_500
Results_Update_500
Results_Evict_500

PR #530, opened 2023-12-04
Optimize LFU maintenance

bitfaster:
Apply [MethodImplOptions.AggressiveOptimization]. This was introduced in .NET Core 3.0, to avoid conditional compile just cast 512 to the enum type which is ignored on earlier versions of .NET.

Dynamic PGO on newer frameworks tends not to optimize the maintenance workload because it doesn't execute on the hot path. This reduces throughput when the cache is contended.


bitfaster:
Not able to measure any gain from this on windows or mac, for .Net6 or .net8. Likely the change for the ConcurrentLruCore struct to be embedded/inlined into the parent class eliminated any gain we might see from this.

[net6base.zip](https://github.com/bitfaster/BitFaster.Caching/files/15342424/net6base.zip)
[net6opt.zip](https://github.com/bitfaster/BitFaster.Caching/files/15342425/net6opt.zip)
[net8base.zip](https://github.com/bitfaster/BitFaster.Caching/files/15342426/net8base.zip)
[net8opt.zip](https://github.com/bitfaster/BitFaster.Caching/files/15342427/net8opt.zip)

PR #531, opened 2023-12-04
Lock LFU nodes when modifying state

bitfaster:


bitfaster:
Fixed in https://github.com/bitfaster/BitFaster.Caching/pull/533

PR #533, opened 2023-12-04
Fix race in LFU maintenance for add-remove-add of the same key

bitfaster:
- Evict removes via KVP, instead of just key. This prevents incorrect removal of nodes added to the dictionary when there are two nodes in the write buffer/LRUs with the same key. For example:
   - Node 1 is added with key x
   - Node 1 is removed, now exists only in write buffer + LRU list
   - Node 2 is added with key x, and is also in the write buffer + LRU list
   - Maintenance previously found node 1, and deleted key x from the dictionary detaching node 2. Delete is now via KVP, so node 2 will not be deleted in this situation.
- Check for removed nodes during `EvictFromMain()`
- `Evict()` marks items as removed/deleted
- Revise Trim logic to use LRU count instead of the dictionary. If TryRemove has been called, the dictionary can contain fewer items than the LRU lists. First call `Maintenance()` to flush buffers and mitigate this.
- Enable node detachment integrity check as part of soak tests

### DIFF
--- a/BitFaster.Caching/Lfu/ConcurrentLfuCore.cs
+++ b/BitFaster.Caching/Lfu/ConcurrentLfuCore.cs
@@ -701,13 +701,17 @@ namespace BitFaster.Caching.Lfu
 
                 if (candidate.node.WasRemoved)
                 {
-                    Evict(candidate.node);
+                    var evictee = candidate.node;
+                    candidate.Next();
+                    Evict(evictee);
                     continue;
                 }
 
                 if (victim.node.WasRemoved)
                 {
-                    Evict(victim.node);
+                    var evictee = victim.node;
+                    victim.Next();
+                    Evict(evictee);
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


BitFaster.Caching/Lfu/ConcurrentLfuCore.cs
```diff
@@ -749,8 +761,11 @@ private bool AdmitCandidate(K candidateKey, K victimKey)
...
         private void Evict(LfuNode<K, V> evictee)
         {
+            evictee.WasRemoved = true;
+            evictee.WasDeleted = true;
+
             this.dictionary.TryRemove(evictee.Key, out var _);
-            evictee.list.Remove(evictee);
+            evictee.list?.Remove(evictee);
             Disposer<V>.Dispose(evictee.Value);
             this.metrics.evictedCount++;
```
bitfaster, line 770: This is now potentially mis-counting.

PR #534, opened 2023-12-05
LRU enumeration should not return expired/removed items

bitfaster:
This fixes a small gap, and introduces an observability difference because TLRU tests previously used to the count to observe eviction. To fix this:

- Remove the abstract base class used for TLRU tests - this is no longer needed since introducing `Duration` and having a single implementation of `TLruLongTicksPolicy`.
- Change the tests to use `ConcurrentTLru`, and verify expiry via the internal queue counters which are part of the public API.

### DIFF
--- a/BitFaster.Caching/Lfu/ConcurrentLfuCore.cs
+++ b/BitFaster.Caching/Lfu/ConcurrentLfuCore.cs
@@ -101,7 +101,7 @@ namespace BitFaster.Caching.Lfu
         }
 
         // No lock count: https://arbel.net/2013/02/03/best-practices-for-using-concurrentdictionary/
-        public int Count => this.dictionary.Skip(0).Count();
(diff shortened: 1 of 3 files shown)


BitFaster.Caching/Lfu/ConcurrentLfuCore.cs
```diff
@@ -101,7 +101,7 @@ public ConcurrentLfuCore(int concurrencyLevel, int capacity, IScheduler schedule
         }
 
         // No lock count: https://arbel.net/2013/02/03/best-practices-for-using-concurrentdictionary/
-        public int Count => this.dictionary.Skip(0).Count();
+        public int Count => this.dictionary.Where(n => !n.Value.WasRemoved).Count();
```
bitfaster, line 104: This doesn't apply - removed _means_ it was removed from the dictionary.

PR #566, opened 2024-05-06
Add new interface to expose 3 new methods without breaking existing clients

wangzq:
The 3 methods were previously added to ICache under netcore3.1 due to default interface implementation is not available for other targets, which makes them inaccessible when using ConcurrentLruBuilder to return ICache, adding a new interface ICache2 with the 3 methods added for other build targets other than netcore3.1 will make them accessible if client cast the returned object to ICache2.


BitFaster.Caching/ICache2.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+namespace BitFaster.Caching
+{
+    /// <summary>
+    /// Represents a generic cache of key/value pairs. This is a new interface with new methods to avoid breaking backward compatibility.
+    /// </summary>
+    /// <typeparam name="K">The type of keys in the cache.</typeparam>
+    /// <typeparam name="V">The type of values in the cache.</typeparam>
+    public interface ICache2<K, V> : ICache<K, V>
```
bitfaster, line 12: I think `ICacheExt` is better, `ICache2` still seems weird to me.

bitfaster (commented):
LGTM

PR #578, opened 2024-05-10
P/Invoke TickCount64 for .NET Standard on Windows

bitfaster:
Note; this introduces a test hole: there are currently no tests that verify .NET standard when it is not running on Windows (we only execute NET4.8 on Windows to test .NET Standard). Since code coverage analysis only runs for .NET Core build targets, this is not observable in the code coverage report.

PR #579, opened 2024-05-10
Test all Duration code paths including .NET Standard

bitfaster:
PR #578 introduced a test hole: `Duration` is not unit tested for non-windows OS on .NET Standard.

- Add a test project running on .NET6 that includes the .NET Standard assembly and test the duration class.
- Modify gate.yml to collect the test results and attach as a file.

The test matrix is like this:

| OS | .NET Compiled | .NET Runtime | Clock | Tested By |
|----|------------------|----------------|--------|-----------|
|Windows| NetStd | Net48 | GetTickCount64 | BitFaster.Caching.UnitTests |
|Windows| NetStd | Net6.0| GetTickCount64 | BitFaster.Caching.UnitTests.Std |
|Windows| Net6.0| Net6.0| Environment.TickCount64 | BitFaster.Caching.UnitTests |
|MacOS| NetStd | Net6.0| Stopwatch.GetTimestamp | BitFaster.Caching.UnitTests.Std |
|MacOS| Net6.0| Net6.0| Stopwatch.GetTimestamp |BitFaster.Caching.UnitTests |
|Linux| NetStd | Net6.0| Stopwatch.GetTimestamp | BitFaster.Caching.UnitTests.Std |
|Linux| Net6.0| Net6.0| Environment.TickCount64 |BitFaster.Caching.UnitTests |

- Windows always uses 64 bit tick count. On .NET standard fall back to P/Invoke.
- MacOS always uses Stopwatch.GetTimestamp.
- Linux has the quirk of using Stopwatch for .NET Std and TickCount64 for .NET6. Benchmarks show this TickCount64 is 3x to 4x faster than StopWatch.GetTimestamp. Without P/Invoke, there is no alternative on .NET Std, and this clock would have anyway been used on Linux even before optimizing MacOS to use Timestamp on .NET6 and Windows to use P/Invoke on Std.


BitFaster.Caching.UnitTests.Std/BitFaster.Caching.UnitTests.Std.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0</TargetFrameworks>
+    <LangVersion>9.0</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="coverlet.msbuild" Version="6.0.0">
```
bitfaster, line 9: If this stomps the code coverage, a simple solution would be to remove coverage collection.

bitfaster:
Verified that if SinceEpoch throws, the test fails on Linux and Mac:

image

BitFaster.Caching.UnitTests.Std/BitFaster.Caching.UnitTests.Std.csproj
bitfaster: Coverage is collected into different directories, so this is not an issue.

bitfaster:
This would have been simpler:

```cs
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public static Duration SinceEpoch()
        {
            if (IsWindows)
            {
#if NETCOREAPP3_0_OR_GREATER
                return new Duration(Environment.TickCount64);
#else
                return new Duration(GetTickCount64());
#endif
            }
            else
            {
                  return new Duration(Stopwatch.GetTimestamp());
            }
        }
```

PR #596, opened 2024-05-26
Mitigate LFU struct tearing (new node)

bitfaster:
When updating a large struct, instead of updating in place, we can instead create a new node. Since nodes are not updated, no tearing can occur.

Time based expiry is a special case that is handled as follows:

- The new `INodePolicy.Clone` method creates the replacement node to insert into the dictionary. For `ExpireAfterPolicy` this avoids calling expiry calculator on create, and we set current expiry to match old node. We also set GetNextInTimeOrder == self, this is a hack so that `ExpireAfterPolicy.OnWrite` knows the node was not created and is currently detached from the `TimerWheel`.
- In `ExpireAfterPolicy.OnWrite`, check if next == self to schedule/reschedule. Schedule will just overwrite next to a valid node.
- Add note to GetNextInTimeOrder to explain

TODO:
- Unit test
- Soak test
- Verify TLFU logic

### DIFF
--- a/BitFaster.Caching/Lfu/ConcurrentLfuCore.cs
+++ b/BitFaster.Caching/Lfu/ConcurrentLfuCore.cs
@@ -371,7 +371,7 @@ namespace BitFaster.Caching.Lfu
                 }
                 else
                 { 
-                    var newNode = policy.Create(key, value);
+                    var newNode = policy.Clone(node, value);
                     if (this.dictionary.TryUpdate(key, newNode, node))
                     { 
                         // This can't be lossy, because we need to attach the new node.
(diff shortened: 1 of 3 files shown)


BitFaster.Caching/Lfu/ConcurrentLfuCore.cs
```diff
@@ -277,686 +277,707 @@ private bool TryGetImpl(K key, [MaybeNullWhen(false)] out V value)
...
+                    this.writeBuffer.TryAdd(node);
+                    TryScheduleDrain();
+                }
+                else
+                { 
+                    var newNode = policy.Create(key, value);
+                    if (this.dictionary.TryUpdate(key, newNode, node))
+                    { 
+                        // This can't be lossy, because we need to attach the new node.
+                        node.WasRemoved = true;
```
bitfaster, line 378: There are 2 issues with time based expiry:

1. `ExpireAfterPolicy.OnWrite` will interpret this as a new node, because `node.GetNextInTimeOrder() == null`. GetExpireAfterUpdate will not be called.
2. `ExpireAfterPolicy.Create `will also call `GetExpireAfterCreate`.

bitfaster:
Superseded by #621

PR #648, opened 2024-11-26
Implement LFU sketch using arm64 intrinsics (redux)

bitfaster:
Below the *BlockAvx benchmark runs on the ARM64 intrinsics, FlatAVX, BlockAVXNotPinned are not implemented. Supersedes https://github.com/bitfaster/BitFaster.Caching/pull/595.

[3a5a72c](https://github.com/bitfaster/BitFaster.Caching/pull/648/commits/3a5a72cd6f35c1add6d0663268ad3b8720b5cae6)
```
BenchmarkDotNet v0.14.0, macOS Sonoma 14.5 (23F79) [Darwin 23.5.0]
Apple M2, 1 CPU, 8 logical and 8 physical cores
.NET SDK 9.0.100
  [Host]   : .NET 6.0.30 (6.0.3024.21525), Arm64 RyuJIT AdvSIMD
  .NET 6.0 : .NET 6.0.30 (6.0.3024.21525), Arm64 RyuJIT AdvSIMD
  .NET 8.0 : .NET 8.0.0 (8.0.23.53103), Arm64 RyuJIT AdvSIMD
  .NET 9.0 : .NET 9.0.0 (9.0.24.52809), Arm64 RyuJIT AdvSIMD
```

BitFaster Caching Benchmarks Lfu SketchFrequency-columnchart

BitFaster Caching Benchmarks Lfu SketchIncrement-columnchart

[3a5a72c](https://github.com/bitfaster/BitFaster.Caching/pull/648/commits/3a5a72cd6f35c1add6d0663268ad3b8720b5cae6)
```
BenchmarkDotNet v0.14.0, Windows 11 (10.0.26100.2314)
Cobalt 100
.NET SDK 9.0.100
  [Host]   : .NET 6.0.36 (6.0.3624.51421), Arm64 RyuJIT AdvSIMD
  .NET 6.0 : .NET 6.0.36 (6.0.3624.51421), Arm64 RyuJIT AdvSIMD
  .NET 8.0 : .NET 8.0.11 (8.0.1124.51707), Arm64 RyuJIT AdvSIMD
  .NET 9.0 : .NET 9.0.0 (9.0.24.52809), Arm64 RyuJIT AdvSIMD
```

BitFaster Caching Benchmarks Lfu SketchIncrement-columnchart

BitFaster Caching Benchmarks Lfu SketchFrequency-columnchart

TODO:
- Measure Mac M series CPU
- Rename AVX benchmarks to Vectorized (assuming it is also worse on M2)
[...]

PR #540, opened 2023-12-10
Use Stopwatch.GetTimestamp on MacOS

bitfaster:
`Stopwatch.GetTimestamp` is slightly faster than `Environment.TickCount64` on MacOS, therefore use the faster implementation to get a small optimization.

This is the time benchmark result from Github actions running on MacOS:

BitFaster Caching Benchmarks TimeBenchmarks ( NET 6 0)

This is the GetOrAdd bench (baseline v2.5.0):

```
BenchmarkDotNet v0.13.12, macOS Sonoma 14.4.1 (23E224) [Darwin 23.4.0]
Apple M1 (Virtual), 1 CPU, 3 logical and 3 physical cores
.NET SDK 8.0.204
  [Host]   : .NET 6.0.29 (6.0.2924.17105), Arm64 RyuJIT AdvSIMD
  .NET 6.0 : .NET 6.0.29 (6.0.2924.17105), Arm64 RyuJIT AdvSIMD
```

BitFaster Caching Benchmarks LruJustGetOrAdd ( NET 6 0)

PR #662, opened 2025-01-26
Update target frameworks

Joy-less:
This pull request removes unsupported target frameworks (.NET Core, .NET 6.0) and simplifies conditional compilation so it supports `NETSTANDARD` (.NET Standard 2.0) and `NET` (.NET 8.0 and .NET 9.0).

### DIFF
--- a/BitFaster.Caching.Benchmarks/DrainBenchmarks.cs
+++ b/BitFaster.Caching.Benchmarks/DrainBenchmarks.cs
@@ -10,7 +10,7 @@ namespace BitFaster.Caching.Benchmarks
     [DisassemblyDiagnoser(printSource: true, maxDepth: 3)]
     [SimpleJob(RuntimeMoniker.Net48)]
 #endif
-    [SimpleJob(RuntimeMoniker.Net60)]
+    [SimpleJob(RuntimeMoniker.Net90)]
     [HideColumns("Job", "Median", "RatioSD", "Alloc Ratio")]
     public class DrainBenchmarks
     {
@@ -187,7 +187,7 @@ namespace BitFaster.Caching.Benchmarks
         public void DrainArray()
         {
             Add();
-#if NETCOREAPP3_1_OR_GREATER
+#if NET
             buffer.DrainTo(output.AsSpan());
 #else
             buffer.DrainTo(new ArraySegment<string>(output));
--- a/BitFaster.Caching.Benchmarks/Lfu/CmSketchNoPin.cs
+++ b/BitFaster.Caching.Benchmarks/Lfu/CmSketchNoPin.cs
@@ -1,10 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 
 
-#if NET6_0_OR_GREATER
+#if NET
 using System.Runtime.Intrinsics;
 using System.Runtime.Intrinsics.Arm;
 using System.Runtime.Intrinsics.X86;
@@ -57,7 +56,7 @@ namespace BitFaster.Caching.Benchmarks.Lfu
         /// <returns>The estimated frequency of the value.</returns>
         public int EstimateFrequency(T value)
         {
-#if NET48
+#if NETSTANDARD
             return EstimateFrequencyStd(value);
 #else
 
@@ -67,7 +66,7 @@ namespace BitFaster.Caching.Benchmarks.Lfu
             {
                 return EstimateFrequencyAvx(value);
             }
-#if NET6_0_OR_GREATER
+#if NET
             else if (isa.IsArm64Supported)
             {
                 return EstimateFrequencyArm(value);
@@ -86,7 +85,7 @@ namespace BitFaster.Caching.Benchmarks.Lfu
         /// <param name="value">The value.</param>
         public void Increment(T value)
         {
-#if NET48
+#if NETSTANDARD
             IncrementStd(value);
 #else
 
@@ -96,7 +95,7 @@ namespace BitFaster.Caching.Benchmarks.Lfu
             {
                 IncrementAvx(value);
             }
-#if NET6_0_OR_GREATER
+#if NET
             else if (isa.IsArm64Supported)
             {
                 IncrementArm(value);
@@ -233,7 +232,7 @@ namespace BitFaster.Caching.Benchmarks.Lfu
             size = (size - (count0 >> 2)) >> 1;
         }
 
-#if NET6_0_OR_GREATER
+#if NET
         private unsafe int EstimateFrequencyAvx(T value)
         {
             int blockHash = Spread(comparer.GetHashCode(value));
@@ -267,7 +266,7 @@ namespace BitFaster.Caching.Benchmarks.Lfu
                     .AsUInt16();
 
                 // set the zeroed high parts of the long value to ushort.Max
-#if NET6_0
+#if NET
                 count = Avx2.Blend(count, Vector128<ushort>.AllBitsSet, 0b10101010);
 #else
                 count = Avx2.Blend(count, Vector128.Create(ushort.MaxValue), 0b10101010);
@@ -333,7 +332,7 @@ namespace BitFaster.Caching.Benchmarks.Lfu
         }
 #endif
 
-#if NET6_0_OR_GREATER
+#if NET
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private unsafe void IncrementArm(T value)
         {
--- a/BitFaster.Caching/Duration.cs
+++ b/BitFaster.Caching/Duration.cs
@@ -26,15 +26,6 @@ namespace BitFaster.Caching
 
         internal static readonly Duration Zero = new Duration(0);
 
-#if NETCOREAPP3_0_OR_GREATER
-        private static readonly bool IsMacOS = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
-#else
-        private static readonly bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
-
-        [DllImport("kernel32")]
-        internal static extern long GetTickCount64();
-#endif
-
         internal Duration(long raw)
         { 
             this.raw = raw; 
@@ -47,25 +38,7 @@ namespace BitFaster.Caching
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static Duration SinceEpoch()
         {
-#if NETCOREAPP3_0_OR_GREATER
-            if (IsMacOS)
-            {
-                return new Duration(Stopwatch.GetTimestamp());
-            }
-            else
-            {
-                return new Duration(Environment.TickCount64);
-            }
-#else
-            if (IsWindows)
-            {
-                return new Duration(GetTickCount64());
-            }
-            else
-            {
-                return new Duration(Stopwatch.GetTimestamp());
-            }
-#endif
+            return new Duration(Stopwatch.GetTimestamp());
         }
 
         /// <summary>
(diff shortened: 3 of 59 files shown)


bitfaster:
Wow, thanks!

I had started to look at this here #638, and planned to go back and effectively make your PR (so I would go in stages update fwks changes, then take each feature as a follow PR). 

I think the gates will fail on your change, I just started a run - my other PR has fixes for that stuff in the build yaml files.

I will take a proper look at this on the weekend.

BitFaster.Caching.Benchmarks/Lfu/CmSketchNoPin.cs
```diff
@@ -57,7 +56,7 @@ public CmSketchNoPin(long maximumSize, IEqualityComparer<T> comparer)
         /// <returns>The estimated frequency of the value.</returns>
         public int EstimateFrequency(T value)
         {
-#if NET48
+#if NETSTANDARD
```
bitfaster, line 59: An executable cannot be .NET standard - this project is an exe and will be compiled as either Framework or Core (3.1/6/8/9). 

So this part has to stay as NET48. You can try this out by running the benchmark on .NET framework.

BitFaster.Caching.Benchmarks/Lfu/CmSketchNoPin.cs
```diff
@@ -86,7 +85,7 @@ public int EstimateFrequency(T value)
         /// <param name="value">The value.</param>
         public void Increment(T value)
         {
-#if NET48
+#if NETSTANDARD
```
bitfaster, line 88: Same as earlier comment on .NET standard an exes.

BitFaster.Caching.Benchmarks/DrainBenchmarks.cs
```diff
@@ -187,7 +187,7 @@ public void Add()
         public void DrainArray()
         {
             Add();
-#if NETCOREAPP3_1_OR_GREATER
+#if NET
```
bitfaster, line 190: This is great - I did not know about the NET directive, much simpler 🙂

BitFaster.Caching.Benchmarks/Lfu/CmSketchNoPin.cs
```diff
@@ -267,7 +266,7 @@ private unsafe int EstimateFrequencyAvx(T value)
                     .AsUInt16();
 
                 // set the zeroed high parts of the long value to ushort.Max
-#if NET6_0
+#if NET
                 count = Avx2.Blend(count, Vector128<ushort>.AllBitsSet, 0b10101010);
 #else
                 count = Avx2.Blend(count, Vector128.Create(ushort.MaxValue), 0b10101010);
```
bitfaster, line 272: This line can now be deleted - it is only required for .NET Core 3.1, because allBitsSet was introduced on .NET6.

BitFaster.Caching/Duration.cs
```diff
@@ -26,15 +26,6 @@ public readonly struct Duration
 
         internal static readonly Duration Zero = new Duration(0);
 
-#if NETCOREAPP3_0_OR_GREATER
```
bitfaster, line 29: There are two optimizations removed by this change:

https://github.com/bitfaster/BitFaster.Caching/pull/578: optimize time calls when running .NET standard on Windows platform.
https://github.com/bitfaster/BitFaster.Caching/pull/540:  optimize time calls on MacOS platform.

BitFaster.Caching/IScopedCache.cs
```diff
@@ -60,8 +60,8 @@ public interface IScopedCache<K, V> : IEnumerable<KeyValuePair<K, Scoped<V>>> wh
         /// the cache.</returns>
         Lifetime<V> ScopedGetOrAdd(K key, Func<K, Scoped<V>> valueFactory);
 
-        // backcompat: remove conditional compile
-#if NETCOREAPP3_0_OR_GREATER
+// backcompat: remove conditional compile
+#if !NETSTANDARD
```
bitfaster, line 64: NET?

BitFaster.Caching/Lfu/CmSketchCore.cs
```diff
@@ -316,7 +311,7 @@ private unsafe int EstimateFrequencyAvx(T value)
                     .AsUInt16();
 
                 // set the zeroed high parts of the long value to ushort.Max
-#if NET6_0_OR_GREATER
+#if NET
                 count = Avx2.Blend(count, Vector128<ushort>.AllBitsSet, 0b10101010);
 #else
                 count = Avx2.Blend(count, Vector128.Create(ushort.MaxValue), 0b10101010);
```
bitfaster, line 317: this line can be deleted - it is for net core 3.1 as mentioned earlier (earlier one is in the copy paste variant for the benchmark).

BitFaster.Caching/Lfu/CmSketchCore.cs
```diff
@@ -340,7 +335,7 @@ private unsafe void IncrementAvx(T value)
             Vector256<ulong> offsetLong = Avx2.PermuteVar8x32(Vector256.Create(index, Vector128<int>.Zero), Vector256.Create(0, 4, 1, 5, 2, 5, 3, 7)).AsUInt64();
             Vector256<long> mask = Avx2.ShiftLeftLogicalVariable(Vector256.Create(0xfL), offsetLong);
 
-#if NET6_0_OR_GREATER
+#if NET
             long* tablePtr = tableAddr;
 #else
             fixed (long* tablePtr = table)
```
bitfaster, line 341: I think this can be removed - also only for net 3.1

BitFaster.Caching.Benchmarks/Lfu/CmSketchNoPin.cs
Joy-less: Understood. Would it be better to use `#if NETFRAMEWORK` here?

Also note: BitFaster.Caching.Benchmarks only has targets for `net8.0` and `net9.0` (previously `net6.0` and `net8.0`) so this wouldn't matter. Are you meaning to add .NET Framework as a target?

BitFaster.Caching/Duration.cs
Joy-less: OK. Normally I am very averse to platform-dependent code, especially invoking external DLLs. Since the benchmarks you ran were on .NET 6.0, there's a good chance this was optimized in .NET 8.0/9.0. This also seems like a micro-optimization since the benchmark diagrams are very very close.
However, if you would still like this code back then I'll reverse this change.

BitFaster.Caching.Benchmarks/Lfu/CmSketchNoPin.cs
bitfaster: `NETFRAMEWORK` is better, assuming benchmark continues to run successfully.

Because Benchmarkdotnet generates a new console application for each benchmark run (via out of process toolchain - to achieve process isolation), it is not necessary to target net48. The benchmarks themselves specify this, e.g. [here](https://github.com/bitfaster/BitFaster.Caching/blob/01805a51bc125d27cd51951a04fdaf04fe466931/BitFaster.Caching.Benchmarks/Lru/LruJustGetOrAdd.cs#L35). Here is the documentation for [toolchains](https://github.com/dotnet/BenchmarkDotNet/blob/master/docs/articles/configs/toolchains.md), note this actually states you must explicitly target all tested frameworks, but in my experience .NET framework behaves differently (as they mention in their docs, because framework is a windows feature, you will always end up running framework benchmarks on the latest version installed on the machine and there is no way to override this). 

I simply eliminated it from the project file, because that was more convenient than having fiddly project level directives to conditionally exclude FWK when building on Linux/MacOS.

BitFaster.Caching/Duration.cs
bitfaster: Please keep in mind the entire purpose of this library is to use micro-optimizations to build the fastest cache on  .NET, across all platforms. If it is possible to make a reproducible optimization with automated functional and performance tests then it is valid for inclusion here, and you will find that all code is tested with very high [coverage](https://coveralls.io/github/bitfaster/BitFaster.Caching?branch=main). This includes platform specific code, such using .NET features like [hardware intrinsics](https://github.com/bitfaster/BitFaster.Caching/pull/648) - platform dependent features have been added to .NET to enable such optimizations. If one platform has a faster way to do something, we can use it and everyone wins.

The first optimization is specifically for .NET Framework - as mentioned in the earlier comment about how the benchmarks are run, you will find this is explicitly tested on .NET Framework. In my day job, I run several services that got a nice gain from this change (the micro-optimization adds up quickly when invoked from tight loops).

The second optimization is related to the API call from .NET to the underlying MacOS operating system. Last I tested this, it is still faster on MacOS Sequoia 15.1.1 (build 24B91). If you are able to demonstrate this has been fixed more recently I would be happy to take the change to remove the code.

Joy-less: Understood, this has been reversed. Thanks for explaining your reasoning.

bitfaster:
I updated the build scripts so that your changes can compile: https://github.com/bitfaster/BitFaster.Caching/pull/670.

Joy-less:
> I updated the build scripts so that your changes can compile: #670.

I merged the changes.

Joy-less:
> Build is failing because deleting framework packages is not backwards compatible (i.e. a breaking change), so this would have to be a major release

Is there going to be a major release soon? What's the plan?

Joy-less:
> Prior to this PR, my thinking was to add a .NET9 build target and implement the .NET9 specific functionality there. Unless there is a dependency on .NET9 APIs (which as of today there is not), there is no advantage to compiling a .NET8 or 9 build target. .NET6 is compatible with those runtimes and have identical performance - hence I did not want to rush into this and introduce regressions which are time consuming to root cause.

.NET 6.0 definitely does [*not*](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-9) have the same performance as .NET 9.0. I assume that what you mean is that a .NET 9.0 project will override the performance of the .NET 6.0 library. Anyways good luck.

bitfaster:
Stephen's post describes optimizations in the JIT at the beginning, before moving onto optimizations in the framework code. In his benchmarks, he compares the results of different versions of the JIT and framework. This is subtly different.

In general, a .NET6 assembly executing on the .NET9 runtime will convert the IL code to assembly code with the .NET9 JIT, run on the .NET9 runtime, and call methods in the .NET9 framework code. Compiling _the same C# code_ to a .NET9 target by itself will not result in any additional optimization because the same IL will be passed to the .NET9 JIT. In the end, identical assembly code will be executed by the CPU in both cases.

To my knowledge there are two possible situations where compiling on .NET9 is faster:
1. The C# language version is updated to C#13, enabling the compiler to bind `lock(object)` to [System.Threading.Lock](https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock) instead of [System.Threading.Monitor](https://learn.microsoft.com/en-us/dotnet/api/system.threading.monitor). Thus, the C# compiler will generate lowered code that invokes a .NET9 specific framework API. Last year I had actually diffed the IL from the different language versions to verify this, but from memory, for BitFaster.Caching, there are no other meaningful differences.
2. The C# code contains regions within `#if NET9_0` that invoke framework APIs that only exist in .NET9, such as the lock API described above.

Because I have no use of `lock(object)` on hot path methods, the first optimization does not materialize any improvement. I know this because I tested it. However, I do have use of explicit `Monitor.TryEnter` in `ConcurrentLfu`, this is an example where the second form of optimization can be used. You can see in the PR I linked my first reply to you has a very small gain due to this: https://github.com/bitfaster/BitFaster.Caching/pull/638.

In summary, when I tried running .NET6 vs .NET9 assemblies in a .NET9 executable, based on this code base, it is not measurably faster. I totally agree .NET9 is faster, but rebuilding the assembly is not the key to unlocking the improvements in this case.

PR #666, opened 2025-02-02
Migrate from FluentAssertions to Shouldly

Joy-less:
As stated in #661, FluentAssertions v8 changed their [license](https://github.com/fluentassertions/fluentassertions/blob/main/LICENSE) from open source (Apache 2.0) to paid costing [$130 per seat](https://xceed.com/products/unit-testing/fluent-assertions) for commercial use.

This pull request migrates the tests in BitFaster.Caching to [Shouldly](https://github.com/shouldly/shouldly), a similar library under the BSD-3-clause license.

Alternatively, you can continue using FluentAssertions without updates, but you must pin it at `[7.0.0]` or `[7.1.0]` to prevent accidental updates.

There was a huge amount to change here so I used this PowerShell script to help me out: [galadril/Shouldly.FromFluentAssertions](https://github.com/galadril/Shouldly.FromFluentAssertions)

### DIFF
--- a/BitFaster.Caching.UnitTests.Std/DurationTests.cs
+++ b/BitFaster.Caching.UnitTests.Std/DurationTests.cs
@@ -2,9 +2,9 @@
 using System.Diagnostics;
 using System.Runtime.InteropServices;
 using BitFaster.Caching.Lru;
-using FluentAssertions;
 using Xunit;
 using Xunit.Abstractions;
+using Shouldly;
 
 namespace BitFaster.Caching.UnitTests.Std
 {
@@ -25,13 +25,11 @@ namespace BitFaster.Caching.UnitTests.Std
             // On .NET Standard, only windows uses TickCount64
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                Duration.SinceEpoch().raw.Should().BeCloseTo(Duration.GetTickCount64(), 15);
+                Duration.SinceEpoch().raw.ShouldBe(Duration.GetTickCount64());
             }
             else
             {
-                // eps is 1/200 of a second
-                ulong eps = (ulong)(Stopwatch.Frequency / 200);
-                Duration.SinceEpoch().raw.Should().BeCloseTo(Stopwatch.GetTimestamp(), eps);
+                Duration.SinceEpoch().raw.ShouldBe(Stopwatch.GetTimestamp());
             }
         }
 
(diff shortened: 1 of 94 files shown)


bitfaster:
Thanks - I hadn't noticed they had changed the terms.

It seems like because this is an open-source project, it is still legal to use v8 here. However, I dislike the change of license terms so would prefer to migrate. I will do some homework on Shouldly.

Joy-less:
> It seems like because this is an open-source project, it is still legal to use v8 here. However, I dislike the change of license terms so would prefer to migrate.

I think it's not legal to use even here because this project is under the MIT license, which technically allows this library to be re-published verbatim under a paid license. It would only be fine if this project was under GPL or another viral license.

BitFaster.Caching.UnitTests/Lru/ConcurrentLruTests.cs
```diff
@@ -108,70 +108,70 @@ public void WhenCapacityIs5HotHasCapacity1AndColdHasCapacity2()
...
 
         [Fact]
         public void WhenItemsAddedKeysContainsTheKeys()
         {
-            lru.Count.Should().Be(0);
+            lru.Count.ShouldBe(0);
             lru.GetOrAdd(1, valueFactory.Create);
             lru.GetOrAdd(2, valueFactory.Create);
-            lru.Keys.Should().BeEquivalentTo(new[] { 1, 2 });
+            lru.Keys.ShouldBe(new[] { 1, 2 });
```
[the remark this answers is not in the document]
bitfaster: There is a discussion here https://github.com/shouldly/shouldly/issues/340

So it should be:

`lru.Keys.ShouldBe(new[] { 1, 2 }, ignoreOrder: true);`

BitFaster.Caching.UnitTests.Std/DurationTests.cs
```diff
@@ -25,13 +25,11 @@ public void SinceEpoch()
             // On .NET Standard, only windows uses TickCount64
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                Duration.SinceEpoch().raw.Should().BeCloseTo(Duration.GetTickCount64(), 15);
+                Duration.SinceEpoch().raw.ShouldBe(Duration.GetTickCount64());
```
bitfaster, line 28: Per https://github.com/shouldly/shouldly/issues/274, looks like you can give the tolerance like this:

`Duration.SinceEpoch().raw.ShouldBe(Duration.GetTickCount64(), 15);`

Not sure if that applies to all data types or just DateTime.

BitFaster.Caching.UnitTests/Lru/ConcurrentLruTests.cs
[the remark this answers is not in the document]
Joy-less: Fixed :)

BitFaster.Caching.UnitTests.Std/DurationTests.cs
Joy-less: Unfortunately, it only applies to DateTime and TimeSpan. Although there is ShouldBeInRange. Another solution could be to use TimeSpan.FromTicks

Joy-less: A little bloated sadly but you can do this:
```cs
TimeSpan.FromTicks(Duration.SinceEpoch().raw).ShouldBe(TimeSpan.FromTicks(Stopwatch.GetTimestamp()), TimeSpan.FromTicks(15));
```

Joy-less: I think I fixed it.

bitfaster: It seems like your current commit fails to build.

The convoluted version converting to TimeSpan is a bit of a mess - will need to think on how that might be cleaned up.

Joy-less: Extension method maybe? I thought you wouldn't want to bloat the project with extension methods.

PR #669, opened 2025-02-05
Use auto-implemented properties

Joy-less:
This eliminates three instances of variable shadowing and slightly simplifies the code.


bitfaster:
Have you verified this produces identical IL and assembly code across .NET Framework and .NET Core?

I have seen in the past that changing to auto properties or changing field order can result in the JIT changing memory layout resulting in regressions that use a lot of time to track down.

Joy-less:
On .NET 6.0:

Before:
image

After:
image

On .NET Standard 2.0:

Before:
image

After:
image

I used `ildasm.exe` on `BitFaster.Caching.dll`. Looks to be the same memory layout to me, just the backing fields have different names.

bitfaster (approved):

PR #686, opened 2025-06-09
Avoid unobserved task exception in AsyncAtomicFactory

advdotnet:
Garbage collecting an `AsyncAtomicFactory` instance can cause an unobserved task exception when a value factory threw.
This is due to a `TaskCompletionSource.SetException` call in the `AsyncAtomicFactory.Initializer` class and its task's exception is never observed.

### DIFF
--- a/.github/workflows/infer.yml
+++ b/.github/workflows/infer.yml
@@ -14,9 +14,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: Setup .NET Core
-      uses: actions/setup-dotnet@v2
+      uses: actions/setup-dotnet@v4
       with:
         dotnet-version: 9.0.x
     - name: Install dependencies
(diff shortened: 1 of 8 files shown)


bitfaster:
Thanks for this - I ran the PR build on your change yesterday and your test fails on .NET6 but not Fwk. I started to look but was paged at work so didn't yet get to understand what is wrong.

advdotnet:
> Thanks for this - I ran the PR build on your change yesterday and your test fails on .NET6 but not Fwk.
> 
> Edit: now I see you pushed more changes - re-running it.

I suspected that it failed in the first run due to another unobserved task exception from other tests. Indeed, there was the same pattern present in the `ScopedAsyncAtomicFactory.Initializer`. After fixing that, I did not encounter any unobserved task exceptions when running the unit tests. However, the test is removed now as it does not provide much value anyway. Let me know if you want it added again, then I could throw and check for a dedicated exception type that will only be used in that test.

bitfaster:
I tried out the scenario using your test - although it is cheesy, I think best approach is to simply await the task to throw the exception inside `AsyncAtomicFactory` (see https://github.com/bitfaster/BitFaster.Caching/pull/688). Using that approach, it is possible to support both avoiding the unobserved task exception and support two or more threads waiting on the same task (the scenario I added the test for).

If you want to merge the await method here with your test and complete this yourself that would be great - I think it is good to have the test running on changes, that way I can't screw it up again.

bitfaster (approved):

bitfaster:
Thanks for this, merged it.

PR #727, opened 2026-01-09
Implement ConcurrentLfu events: attempt 3

bitfaster:
This PR implements events for `ConcurrentLfu` as a switchable policy. When disabled all event code is fully elided at runtime by the JIT compiler.

Events are considered perf critical in `ConcurrentLfu` because the ItemRemoved logic is invoked as part of the maintenance cycle, this introduces overhead even when there are no events registered. The Maintenance method latency determines cache throughput at the limit, so any overhead here is not desired. Later, these events could be captured in a list and processed asynchronously via the scheduler.

In this implementation, calling `TryRemove` defers event processing to the maintenance cycle (thus `TryRemove` and policy based eviction behave the same), whereas `TryUpdate` executes the event handler immediately when `TryUpdate` is called. 

This PR was produced with Claude Code using the following prompts:

Prompt 1:

> Implement the ItemRemoved and ItemUpdated events for ConcurrentLfu. Do not modify the LFU builder classes yet. Follow the pattern of ITelemetryPolicy in ConcurrentLru, but do not incorporate the cache hit metrics because they are already implemented in ConcurrentLfu. Instead of ITelemetryPolicy, for ConcurrentLfu define a new interface IEventPolicy with the ItemRemoved and ItemUpdated events. IEventPolicy should be implemented as a struct to enable the JIT to elide unused code when events are not yet enabled.

Prompt 2:

> Implement the same IEventPolicy changes made in ConcurrentLfu in the ConcurrentTLfu class.

Prompt 3:

> Fix INodePolicy and the classes derived from it to accept the event policy generic parameter.

Fixed manually: reviewed code to determine correct approach to distinguish between evicted vs removed.

Fixed manually: test build and run.

Prompt 4:

> Add unit tests to ConcurrentLfuTests for the ItemRemoved and ItemUpdated events. For ItemRemoved, explicitly test
  different removed reasons. When calling TryRemove, reason should be Removed. When the cache reaches bounded size and
  the policy evicts an item, reason should be evicted. When calling Trim and Clear, reason should be set accordingly.

Prompt 5:

> For the ConcurrentLfu and ConcurrentTLfu event properties, replicate the Proxy inner class from ConcurrentLruCore that enables the telemetry policy mutable struct to function correctly. 

Prompt 6:

> In ConcurrentLfuCore, modify the Evict method to accept an ItemRemovedReason argument. When the Maintenance method is invoked from Trim, ItemRemovedReason should be trimmed. When Maintenance is invoked from Clear, ItemRemovedReason should be cleared. In all other cases, Evict should use ItemRemovedReason Evicted. Do not introduce a class level flag to track ItemRemovedReason, instead propagate the reason as a method argument from the call site where the reason is known.  The existing unit tests cover each of these cases, and the tests for cleared and trimmed are failing. This change should fix those two test cases.

Prompt 7:

> Add explicit unit tests for the NoEventPolicy and EventPolicy policy classes. The tests should exercise all methods and properties, similar to the TelemetryPolicyTests used to test the TelemetryPolicy class.

Prompt 8:
> Replicate the tests for ConcurrentLfu in ConcurrentLfuTests for item updated and item removed events for
  ConcurrentTLfu in ConcurrentTLfuTests.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,11 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(dotnet test:*)",
+      "Bash(dotnet build:*)",
+      "Bash(ls:*)",
+      "Bash(find:*)",
+      "Bash(dotnet format:*)"
+    ]
+  }
+}
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -5,7 +5,8 @@
       "Bash(dotnet build:*)",
       "Bash(ls:*)",
       "Bash(find:*)",
-      "Bash(dotnet format:*)"
+      "Bash(dotnet format:*)",
+      "Bash(dotnet run:*)"
     ]
   }
 }
--- a/.claude/skills/bench-fast/skill.md
+++ b/.claude/skills/bench-fast/skill.md
@@ -1,0 +1,34 @@
+---
+name: bench-fast
+description: Run a benchmark with minimal iterations to generate assembly code via BenchmarkDotNet's DisassemblyDiagnoser. Use to generate one big assembly code file per benchmark. 
+---
+
+## Usage
+
+```
+/bench-fast <BenchmarkName> [<Runtimes>]
+```
+
+## Arguments
+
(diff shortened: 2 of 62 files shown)


bitfaster:
@copilot resolve the merge conflicts

bitfaster:
@copilot fix all merge conflicts

PR #762, opened 2026-04-14
Additional soak test diagnostics

bitfaster:


bitfaster:
Theory:
- This started after fixing background scheduler dispose in https://github.com/bitfaster/BitFaster.Caching/pull/750
- It does not look safe to run maintenance back to back. There is no guard against no candidate returned from window eviction.
- In that case running maintenance multiple times should repro.

PR #773, opened 2026-04-16
Inline GetAlternateLookup

bitfaster:
Inlining GetAlternateLookup eliminates the heap allocation for the AlternateLookup struct on .NET 10:

| Method               | Runtime   | Mean     | Ratio | Gen0   | Allocated |
|--------------------- |---------- |---------:|------:|-------:|----------:|
| GetAlternateInline   | .NET 10.0 | 42.73 ns |  0.77 |      - |         - |
| GetAlternateNoInline | .NET 10.0 | 52.94 ns |  0.96 | 0.0092 |      40 B |
| GetAlternateInline   | .NET 9.0  | 55.35 ns |  1.00 | 0.0092 |      40 B |
| GetAlternateNoInline | .NET 9.0  | 56.13 ns |  1.01 | 0.0092 |      40 B |

From https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-10/runtime

> ### Inlining improvements
> Various inlining improvements have been made in .NET 10.
>
> The JIT can now inline methods that become eligible for devirtualization due to previous inlining. This improvement allows the JIT to uncover more optimization opportunities, such as further inlining and devirtualization.
>
> ...
>
> #### Return types
> During inlining, the JIT now updates the type of temporary variables that hold return values. If all return sites in a callee yield the same type, this precise type information is used to devirtualize subsequent calls. This enhancement complements the improvements in late devirtualization and array enumeration de-abstraction.

### DIFF
--- a/BitFaster.Caching.Benchmarks/AltBenchmark.cs
+++ b/BitFaster.Caching.Benchmarks/AltBenchmark.cs
@@ -12,8 +12,11 @@ using BitFaster.Caching.Scheduler;
 
 namespace BitFaster.Caching.Benchmarks
 {
+    // The get alternate lookup family of methods return an interface, while under the hood they return a struct.
+    // This benchmark verifies whether the JIT can devirtualize the return type to avoid boxing the struct.
+    // It is not currently possible to avoid allocs for TryGet.
     [MemoryDiagnoser]
-    [HideColumns("Job", "Median", "RatioSD")]
+    [HideColumns("Job", "Median", "RatioSD", "Error", "StdDev")]
     public class AltBenchmark
     {
         private static readonly ConcurrentLru<string, int> concurrentLru = new ConcurrentLru<string, int>(8, 9, EqualityComparer<string>.Default);
@@ -28,9 +31,8 @@ namespace BitFaster.Caching.Benchmarks
         }
 
 #if NET9_0_OR_GREATER
-
         [Benchmark]
-        public int LruGetAlternateInline()
+        public int LruGetAlternate()
         {
             var alt = concurrentLru.GetAlternateLookup<ReadOnlySpan<char>>();
             alt.TryGet("1", out int value);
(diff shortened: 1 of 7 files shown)


bitfaster:
inlining TryGet does not eliminate allocs.

An alternative is to make TryGet an extension method, and in the extension method use the ICache comparer to do the check, then call GetAltLookup where the inlining trick works.

bitfaster:
Tested both LRU and LFU end to end:

| Method                | Runtime   | Mean      | Gen0   | Allocated |
|---------------------- |---------- |----------:|-------:|----------:|
| LruGetAlternateInline | .NET 10.0 |  42.69 ns |      - |         - |
| LfuGetAlternateInline | .NET 10.0 |  76.70 ns |      - |         - |
| LruGetAlternateInline | .NET 9.0  |  53.58 ns | 0.0092 |      40 B |
| LfuGetAlternateInline | .NET 9.0  | 127.16 ns | 0.0353 |     152 B |

bitfaster:
Also tried this (including forcing the return type to always be the same struct), but does not devirtualize on .NET 10:

```cs
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public bool TryGetAlternateLookup<TAlternateKey>([MaybeNullWhen(false)] out IAlternateLookup<TAlternateKey, K, V> lookup)
            where TAlternateKey : notnull, allows ref struct
        {
            AlternateLookup<TAlternateKey> alt = AlternateLookup<TAlternateKey>.Empty;
            if (this.dictionary.IsCompatibleKey<TAlternateKey, K, I>())
            {
                alt = new AlternateLookup<TAlternateKey>(this);
                lookup = alt;
                return true;
            }

            lookup = alt;
            return false;
        }

        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public (bool, IAlternateLookup<TAlternateKey, K, V>)TryGetAlternateLookup2<TAlternateKey>()
    where TAlternateKey : notnull, allows ref struct
        {
            if (this.dictionary.IsCompatibleKey<TAlternateKey, K, I>())
            {
                return (true, new AlternateLookup<TAlternateKey>(this));
            }

            return (false, AlternateLookup<TAlternateKey>.Empty);
        }
```

PR #791, opened 2026-04-28
Allow ref struct for TArg in ICache GetOrAdd methods

### LINKED ISSUE #774: Allow ref struct for TArg type parameter in GetOrAdd methods
The `GetOrAdd<TArg>` methods currently do not allow `TArg` to be a ref struct. It would be great if the `allows ref struct` anti-constraint could be added for `TArg` to these methods (on .NET runtimes 9.0 and later) so we can pass a `Span<T>` parameter.

bitfaster:


bitfaster:
@copilot propose some options to fix the build error. Fix should minimize code change and consider performance.

PR #797, opened 2026-05-01
Add support for WithEvents to ConcurrentLfuBuilder

bitfaster:
- Builder now supports `WithEvents`. Previously, `ConcurrentLfu` did not have events, so this preserves existing behavior for the builder.
- Builder returns internal class `FastConcurrentLfu` which uses the `NoEventsPolicy`. It supports both access order and time based expiry via generic parameters. This means we can have a single class for both expiry modes. Since this is internal, user cannot see the ugly generics. We can later consider providing public `FastConcurrentLfu` and `FastConcurrentTLfu` and clean up the generics.
- Implemented a test class base with derived test classes to verify basic functionality for access order, expire after access, expire after write and expire after policies.


BitFaster.Caching.UnitTests/Lfu/ConcurrentLfuSoakTests.cs
```diff
@@ -472,7 +472,7 @@ private void Checker(ICache<int, Guid> cache, CancellationTokenSource source)
         private ConcurrentLfu<int, string> CreateWithBackgroundScheduler()
         {
             var scheduler = new BackgroundThreadScheduler();
-            return new ConcurrentLfuBuilder<int, string>().WithCapacity(9).WithScheduler(scheduler).Build() as ConcurrentLfu<int, string>;
+            return new ConcurrentLfuBuilder<int, string>().WithCapacity(9).WithScheduler(scheduler).WithEvents().Build() as ConcurrentLfu<int, string>;
```
bitfaster: why is this needed?

bitfaster:
@copilot fix the failing tests. Do not make any changes except for fixing the tests.

PR #814, opened 2026-06-22
Add weight-based eviction to ConcurrentLfu

bitfaster:
## Summary

Adds opt-in **weight-based eviction** to `ConcurrentLfu`, faithfully matching Caffeine's weighted W-TinyLFU. A user supplies an `IWeigher<K, V>`; the cache is then bounded by **total weight** instead of item count.

```csharp
var lfu = new ConcurrentLfuBuilder<string, byte[]>()
    .WithCapacity(1_000_000)              // maximum total weight
    .WithWeigher(new ByteArrayWeigher())  // IWeigher<K,V>.Weigh => value.Length
    .Build();
```

`WithWeigher` composes with `WithEvents` and `WithExpireAfterWrite`/`WithExpireAfterAccess`/`WithExpireAfter`.

## Design

- **No new generic.** Weighting is carried by the existing `INodePolicy` slot via two new policy structs (`WeightedAccessOrderPolicy`, `WeightedExpireAfterPolicy`). `ConcurrentLfuCore`/`TimerWheel` arities are unchanged.
- **Zero cost to the default (unweighted) path.** Weighted branches are selected by `static readonly bool IsWeighted = default(P).IsWeighted` and JIT-elided in the count case. Weight is stored on weighted node subclasses, so unweighted node layout is untouched.
- **`policyWeight`-authoritative accounting** maintains the invariant `weightedSize == Σ policyWeight`, which is robust to BitFaster's buffered/duplicated writes.
- Ports Caffeine's weighted `evictFromWindow`/`evictFromMain` (zero-weight skips, oversize-candidate eviction, frequency admission with the `>=6`/`1-in-128` anti-hashflood jitter), oversize-on-add, conditional weight-change handling, zero-weight survival, and the absolute-weight hill-climb (`increase`/`decreaseWindow` moving whole nodes within quota).
- `IBoundedPolicy.Capacity` stays `int` (the weight budget); internal accounting is `long`. With a weigher, zero/low-weight entries mean `Count` can exceed `Capacity` (by design).

## Scope

In: `ConcurrentLfu` (with/without events) and `ConcurrentTLfu` (weight composed with time expiry). Deferred: `FastConcurrentLfu` standalone exposure and the scoped/atomic/async builder wrappers.

## Testing

- Full suite: **1563 pass**, no regressions; builds on all TFMs (netstandard2.0, netcoreapp3.1, net6.0, net10.0, net48).
- New coverage: weighted accounting/eviction/oversize/zero-weight/weight-change/promotion invariants, the hill-climb adaptation, all builder combinations, the shared cache suite run against the weighted+events wrapper, and a 4-thread concurrency soak that reuses the integrity checker and asserts the weighted-size invariant under load.
- A rubber-duck review caught a critical bug (probation→protected read promotion was using the unweighted path, corrupting `mainProtectedWeightedSize`) — fixed and regression-tested — plus three weighted+expiry hazards (deschedule-on-remove, skip `AfterWrite` on evict, idempotent discount).

## Note

`TimerWheelTests.WhenRescheduledLaterNodeIsMoved` is a pre-existing, date/wall-clock-dependent test (uses real `Duration.SinceEpoch()`); it fails on a clean `main` independent of this change.

## Draft

Opened as a draft for review of the approach and API shape before finalizing.

### DIFF
--- a/BitFaster.Caching.UnitTests/Lfu/ConcurrentLfuBuilderTests.cs
+++ b/BitFaster.Caching.UnitTests/Lfu/ConcurrentLfuBuilderTests.cs
@@ -577,25 +577,45 @@ namespace BitFaster.Caching.UnitTests.Lfu
         }
 
         [Fact]
-        public void WithWeigherAndEventsThrows()
+        public void WithWeigherAndEventsBuildsWeightedCacheWithEvents()
         {
-            var builder = new ConcurrentLfuBuilder<int, int>()
+            ICache<int, int> weighted = new ConcurrentLfuBuilder<int, int>()
+                .WithCapacity(100)
                 .WithWeigher(new IntValueWeigher())
-                .WithEvents();
+                .WithEvents()
+                .Build();
 
-            Action act = () => builder.Build();
-            act.Should().Throw<InvalidOperationException>();
+            weighted.Should().BeAssignableTo<WeightedConcurrentLfu<int, int, WeightedAccessOrderNode<int, int>, WeightedAccessOrderPolicy<int, int, EventPolicy<int, int>>>>();
+            weighted.Events.HasValue.Should().BeTrue();
         }
 
         [Fact]
-        public void WithWeigherAndExpireAfterWriteThrows()
+        public void WithWeigherAndExpireAfterWriteBuildsWeightedCacheWithExpiry()
         {
-            var builder = new ConcurrentLfuBuilder<int, int>()
+            ICache<int, int> weighted = new ConcurrentLfuBuilder<int, int>()
+                .WithCapacity(100)
                 .WithWeigher(new IntValueWeigher())
-                .WithExpireAfterWrite(TimeSpan.FromSeconds(1));
+                .WithExpireAfterWrite(TimeSpan.FromSeconds(1))
+                .Build();
 
-            Action act = () => builder.Build();
-            act.Should().Throw<InvalidOperationException>();
+            weighted.Should().BeAssignableTo<FastConcurrentLfu<int, int, WeightedTimeOrderNode<int, int>, WeightedExpireAfterPolicy<int, int, NoEventPolicy<int, int>>>>();
+            weighted.Policy.ExpireAfterWrite.HasValue.Should().BeTrue();
+            weighted.Policy.ExpireAfterWrite.Value.TimeToLive.Should().Be(TimeSpan.FromSeconds(1));
+        }
+
+        [Fact]
+        public void WithWeigherAndExpireAfterWriteAndEventsBuildsWeightedCache()
+        {
+            ICache<int, int> weighted = new ConcurrentLfuBuilder<int, int>()
+                .WithCapacity(100)
+                .WithWeigher(new IntValueWeigher())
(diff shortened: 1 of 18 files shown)


ben-manes:
fwiw, this [paper](https://dl.acm.org/doi/10.1145/3507920?cid=99659224047) argued for a different weighted admission scheme than Caffeine's (compares an accumulation). 

My concerns with the approach was that under concurrency this becomes harder to reason about because the cache's state is constantly changing. There is still a need for an evaluation loop to handle races, e.g. the entry's weight might change mid evaluation or it could be explicitly removed. Since weight-based caches are less common for on-heap application caches (hard to size objects, smaller, use less memory) it seemed like an optimization that could be deferred. The approach would make more sense for persisted or dedicated remote cache servers which might hold hundreds of millions of entries, but also have vastly different needs and design goals. You might find something interesting to borrow as I liked the idea conceptually but there was a marginal hit rate increase and no real world advocate.

bitfaster:
Interesting, thanks for this. If I understood, today Caffeine follows the Implicit Victims (IV) scheme, where admit only compares one candidate and victim, then the evict from main loop compares the frequency of the candidate/victim pair removing one by one until the cache is within bounds. A large candidate can evict multiple smaller victims. 

With Aggregated Victims (AV), the cumulative frequency of the victims is considered, such that the candidate frequency must be greater than the sum of victims. Intuitively, this makes sense to me - if there are many small items, evicting them could cause many cache misses that outweigh the benefit of caching the single large item.

The practical implication of AV is that the maintenance loop now needs another buffer to store the aggregated victims, and if there can be zero weight items the buffer can be the same size as the cache itself in the worst case. I didn't think too hard but it seems like items explicitly removed or items that change weight could simply be ignored, i.e. just assume those things didn't happen. Because the maintenance loop is anyway like a snapshot this would be no different.

If the real-world improvement is more like noise, I would tend to agree it's not worth complicating the code. I have a lot of work to do on this PR to factor the code cleanly - I was curious if Opus 4.8 could do it. Like most things the answer is technically yes, but it's not great if you are an old-fashioned engineer that reads code.

ben-manes:
yep, IV was just the simple obvious path with so much other complexity we have to wade through and focus on. The AV intuitively seems better and not adding much more complexity, since we already have to handle races (resurrect) so as you said its likely that little changes. AI helps make it a lot easier to add that bit of extra complexity, the various test cases, and thoroughly verify it is a hit rate improvement. The marginal but real gain wasn't enough in the old-fashioned engineering style to motivate me to do it, mostly because sifting through dozens of trace results manually was so time consuming.

iirc, one minor quirk in the paper is that they used an admission policy of `greater-than-or-equal-to` instead of `greater-than`, which allowed new arrivals with the same score to be admitted. Thus it would fail a Lirs loop trace by acting as an LRU where an MRU is needed to protect the working set. iirc this helped their workloads, so the real gains might be lower.

I think its worth keeping in a todo backlog which is why its on my mind to get to someday. I think AI makes it much more approachable. Shouldn't be a priority and just an enhancement someday if either of us get curious enough to work on it.

